PR #2, opened 2017-08-31
Update README.md for installation

mhemmings:
`go get` will get all dependancies, so there is no need for a second step

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,14 +12,6 @@ Install the SDK using the following command:
 go get -u github.com/aws/aws-xray-sdk-go
 ```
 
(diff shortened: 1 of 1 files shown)


dlsniper:
The problem is that there are no sources in the root of the package ` github.com/aws/aws-xray-sdk-go ` so ` go get ` will not follow the packages it contains automatically. To do so, either see my commit that references this PR or add `  go get -u github.com/aws/aws-xray-sdk-go/... ` to the instructions.

mhemmings:
Ah OK I see, that's an awkward one. But I agree dfb09e1 seems to be the suitable way round it.

PR #3, opened 2017-08-31
use lower case strings for traceid

DaveBlooman:
My use case is to have a something create the Trace ID before reaching the Go app, however that service only generates uppercase strings and Xray only accepts lowercase strings.

### DIFF
--- a/xray/handler.go
+++ b/xray/handler.go
@@ -102,7 +102,7 @@ func Handler(sn SegmentNamer, h http.Handler) http.Handler {
 
 		trace := parseHeaders(r.Header)
 		if trace["Root"] != "" {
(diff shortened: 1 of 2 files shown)


abraithwaite:
This looks like a breaking change.  Probably worth adding a test and considering the impacts to other consumers of the package.

jamesdbowman:
Hi Dave,

Thanks for your pull request. Unfortunately, this is currently a limitation of the AWS X-Ray service. Making this change in the SDK would limit future revisions to the trace ID, if they were to occur.

We will consider a backwards-compatible change to our service to accept non-lowercase trace IDs in the future.

Closing this for now.

PR #7, opened 2017-09-03
Overall cleanup and readability improvements

dlsniper:
Hi all,

Thank you for releasing the SDK.

I've made a few changes here and there, and I'm happy to do discuss them.

There's a breaking change as well, renaming ` HttpSubsegments  ` to ` HTTPSubsegments ` in order to confirm with the ` golint ` initialism.

I plan to have further changes as well.

Refactor out as much as possible of the package level variables as well as the calls to ` os.Getenv() `, especially in the ` xray/handler.go ` file.

A final thing that I would love to see changed would be moving everything from ` github.com/aws/aws-xray-sdk-go/xray ` to ` github.com/aws/aws-xray-sdk-go ` as the package can still be named ` xray ` but it drops the extra level of indirection, which doesn't seem to be needed as in this case.

Please let me know how to best make sure that these changes get merged in before 1.0 release.

Looking forward for your reply.

Issue #1, opened 2017-08-31
README suggestion


alexbilbie:
After some experimentation I've been able to successfully integrate this into my application, thank you!

After the "Configuration" section and before the "Capture" section in the README's Quick Start can I recommend adding a simple example of how to start a segment and subsegment for those developers who aren't working with HTTPS based services:

```go
// Start a segment:
ctx, seg := xray.BeginSegment(context.Background(), "my-service-name")

ctx, subSeg := xray.BeginSubsegment(ctx, "Some unit of work")
// ...
// some unit of work here
// ...
subSeg.Close(nil)

// Close the segment
seg.Close(nil)
```

luluzhao:
@alexbilbie , we released a new version of SDK which contains your suggestions. Thank you for your suggestion.

PR #9, opened 2017-09-03
Improve README examples

dlsniper:
This PR adds some improved examples of how to use XRay from Go.

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,16 +9,8 @@ The AWS X-Ray SDK for Go is compatible with Go 1.7 and above.
 Install the SDK using the following command:
 
 ```
-go get -u github.com/aws/aws-xray-sdk-go
+go get -u github.com/aws/aws-xray-sdk-go/...
 ```
-
-Some external dependencies are required. They may be installed using:
-
-```
-glide install
-```
-
-Or by individually calling `go get` for each of the required dependencies.
 
 ## Getting Help
 
--- a/README.md
+++ b/README.md
@@ -57,7 +57,51 @@ func init() {
 }
 ```
 
+** Creating a trace **
+
+```go
+package main
+
+import (
+	"context"
+
+	"github.com/aws/aws-xray-sdk-go/xray"
+
+	// Importing the plugins enables collection of AWS resource information at runtime
+	_ "github.com/aws/aws-xray-sdk-go/plugins/ec2"
+	_ "github.com/aws/aws-xray-sdk-go/plugins/beanstalk"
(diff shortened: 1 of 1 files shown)


dlsniper:
I've used @alexbilbie's example from #1 to add the example of using a custom segment / subsegment as I was basically too lazy to type it on my own. Not sure how this works out for copyright, please advise.

dlsniper:
Once #7 lands, errors can be added as well.

luluzhao:
Hi @dlsniper, thank you for your suggestions about refactoring README. I gonna start modify it based on your suggestions. I agree with #1 , however, for #2 , I gonna keep glide section since some of customer might use it and what is more, "go get -u github.com/aws/aws-xray-sdk-go/..." only install non-test packages and it customer want to run unit test. They still need to install dependencies for unit-test by themselves.

dlsniper:
> I gonna keep glide section since some of customer might use it and what is more, "go get -u github.com/aws/aws-xray-sdk-go/..." only install non-test packages and it customer want to run unit test. They still need to install dependencies for unit-test by themselves.

You can add ` -t ` to ` go get ` for getting the testing dependencies. I understand that you prefer having them at a specific version, but this won't work too well for various dependency managers if the user runs ` glide install `. I agree that overall this is not a good solution but sadly that's the current state of package management in Go.

Feel free to send the changes in the branch, it's open to changes form maintainers.

README.md
```diff
@@ -83,17 +85,40 @@ func criticalSection(ctx context.Context) {
 **HTTP Handler**
 
 ```go
-func main() {
-  http.Handle("/", xray.Handler(xray.NewFixedSegmentNamer("myApp"), http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+package main
```
luluzhao: From my opinion, I don't think it's a good idea to add library information into every example. Lots of code are just repeated.

dlsniper: The problem that I had (recently when I started to need to integrate x-ray again) is that I couldn't find these easy to do. Maybe a better way would be to have a few [testable examples](https://blog.golang.org/examples) instead? I'm happy to redo the PR if so.

luluzhao:
@dlsniper , I am sorry we get back to you after a long time. We updated our README and took part of your suggestion. We decide to keep `glide install` section since some customer may use it. And also I think it's not a good idea to write down library in every example since lots of code are just repeated again and again.

README.md
luluzhao: In README configuration section, there is a part to tell customer to import xray library. Have you take a look at that? Could you please tell me which example you were getting stuck into?

PR #21, opened 2017-09-26
Adds NewHTTPSubsegments

jeromegn:
It's currently impossible to use a custom `httptrace.ClientTrace` with `*xray.HTTPSubsegments` because opCtx is not exported.

This PR adds a `func NewHTTPSubsegments(opCtx context.Context) *HTTPSubsegments` to allow people with already existing `httptrace.ClientTrace` to plug in AWS X-Ray.

### DIFF
--- a/xray/httptrace.go
+++ b/xray/httptrace.go
@@ -25,6 +25,12 @@ type HTTPSubsegments struct {
 	responseCtx context.Context
 }
 
+// NewHTTPSubsegments creates a new HTTPSubsegments to use in
+// httptrace.ClientTrace functions
+func NewHTTPSubsegments(opCtx context.Context) *HTTPSubsegments {
+	return &HTTPSubsegments{opCtx: opCtx}
+}
+
 // GetConn begins a connect subsegment if the HTTP operation
 // subsegment is still in progress.
 func (xt *HTTPSubsegments) GetConn(hostPort string) {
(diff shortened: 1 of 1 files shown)


xray/httptrace.go
```diff
@@ -25,6 +25,12 @@ type HTTPSubsegments struct {
 	responseCtx context.Context
 }
 
+// NewHTTPSubsegments creates a new HTTPSubsegments to use in
+// httptrace.ClientTrace functions
+func NewHTTPSubsegments(opCtx context.Context) *HTTPSubsegments {
```
luluzhao: Could you tell me more about this use case? It seems like it just uses a different way to pass opCtx params.

jeromegn: Sure.

You can't assign an unexported field from outside a package.

Example:

```go
// $GOPATH/src/github.com/superfly/unexportedtest/somepkg/somepkg.go
package somepkg

type ExportedStruct struct {
	aField string
}

// $GOPATH/src/github.com/superfly/unexportedtest/main.go
package main

import "github.com/superfly/unexportedtest/somepkg"

func main() {
	some := somepkg.ExportedStruct{aField: "hello"}
}
```

Result:

```
$ go build .
# github.com/superfly/unexportedtest
./main.go:6:39: unknown field 'aField' in struct literal of type somepkg.ExportedStruct
```

I wanted to use `HTTPSubegments` from my own package but I couldn't initialize it with a context since the field `opCtx` is not exported (capitalized).

Creating this `NewHTTPSubsegments` function fixes that issue.

luluzhao: Thank you for your example and now I understand your use case. Basically you want use `HTTPSubsegment` struct but every item under this struct is lowercase so that you cannot initialize. I will merge this fix and thank you again for your fix.

luluzhao:
@jeromegn , thank you for your fix and I already merge it into our branch. Close this thread.

PR #30, opened 2017-11-09
Use `req.URL.Host` for a segment name if `req.Host` is not defined

pottava:
It is not difficult to set a request host normally. But in some cases it is VERY difficult. When I want to use OpenAPI & go-swagger to generate source code for instance, http.Request will be created inside the library and the host is not configured. As a result, x-ray daemon would NOT send the request to AWS because the segment name is not set properly. This PR will fix the issue by using r.URL.Host if the r.Host is not defined.


luluzhao:
@pottava , thank you for your fix and we merge it into your branch.

PR #32, opened 2017-12-20
Fix Data Races

raymolin:
Fixes data race between Context.Done and Segment.Close()

**Scenario**: When a segment is created, a goroutine is also spawned which listens on the Context.Done channel and flushes the segment if it hasn't been already.  Race condition is created when the main application is trying to close the segment and the Context is cancelled at the same time.

**Issue**: Segment.Close() locks the segment as it modifies a condition to determine whether the segment should be flushed.  However, the goroutine doesn't take a lock trying to access that information.

**Solution**: Refactor flush method itself to not care about locking, and instead have the callers keep track of the locking state.

Refactor injectable logger:
- Rip out dependency on cihub/seelog and instead create an interface in case users of the library want to override the naive implementation using the stdlib "log" library

### DIFF
--- a/logger/log.go
+++ b/logger/log.go
@@ -1,0 +1,33 @@
+package logger
+
+func Debug(msg string) {
+	internalLogger.Debug(msg)
+}
+
+func Debugf(format string, args ...interface{}) {
+	internalLogger.Debugf(format, args)
+}
+
+func Info(msg string) {
+	internalLogger.Info(msg)
+}
+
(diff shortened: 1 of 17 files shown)


logger/log.go
```diff
@@ -0,0 +1,33 @@
+package logger
```
luluzhao: We will have a new logger interface in future to allow customer to inject io.writer. What is more, we will also support logger per instance instead of just support global logger.

PR #33, opened 2018-01-10
Allow nil segments to noop instead of panic

dlsniper:
I'm using X-Ray like this:

The initialization portion:

```go
// XRayMissingCtxNOOP makes the XRay configuration not panic if the XRay context is missing
type XRayMissingCtxNOOP struct{}

func (XRayMissingCtxNOOP) ContextMissing(v interface{}) {}

// in the app initialization part:
func someFunc() {
	xray.Configure(xray.Config{
		DaemonAddr:     "127.0.0.1:2000",
		LogLevel:       "warn",
		ServiceVersion: serviceVersion,
		ContextMissingStrategy: XRayMissingCtxNOOP{},
	})
}

// XRayHandler allows for the request tracing
func (a *App) XRayHandler(next http.HandlerFunc) http.HandlerFunc {
	if a.fsn != nil {
		return func(w http.ResponseWriter, r *http.Request) {
			if a.shouldTrace() {
				xray.Handler(a.fsn, http.HandlerFunc(next)).ServeHTTP(w, r)
			} else {
				next(w, r)
			}
		}
	}

	return next
}
```

The handlers all are:

```
func someHandler() {
	_, subSeg := xray.BeginSubsegment(r.Context(), "http.guaranteedSend")
	var err error
	subSeg.Close(err)
	// code...
}
```

But the problem is that whenever I don't inject the xray context in it, ` Close() ` will panic.

As such, this PR no-ops the operations that are used on segments.

If there's a better way to address this, please let me know.

### DIFF
--- a/xray/segment.go
+++ b/xray/segment.go
@@ -127,7 +127,9 @@ func NewSegmentFromHeader(ctx context.Context, name string, h *header.Header) (c
 
 // Close a segment.
 func (seg *Segment) Close(err error) {
-
+	if seg == nil {
+		return
+	}
 	seg.Lock()
 	if seg.parent != nil {
 		log.Tracef("Closing subsegment named %s", seg.Name)
@@ -147,6 +149,9 @@ func (seg *Segment) Close(err error) {
 
 // RemoveSubsegment removes a subsegment child from a segment or subsegment.
 func (seg *Segment) RemoveSubsegment(remove *Segment) bool {
+	if seg == nil {
+		return true
+	}
 	seg.Lock()
 	defer seg.Unlock()
 
@@ -216,6 +221,9 @@ func (seg *Segment) addPlugin(metadata *plugins.PluginMetadata) {
 
 // AddAnnotation allows adding an annotation to the segment.
 func (seg *Segment) AddAnnotation(key string, value interface{}) error {
+	if seg == nil {
+		return nil
+	}
 	switch value.(type) {
 	case bool, int, uint, float32, float64, string:
 	default:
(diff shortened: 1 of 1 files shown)


luluzhao:
@dlsniper , thank you for reaching out to us about your issue. For my understanding, you problem is when you try to close a nil subsegment, program will start panic since we didn't do nil check. However, I am wondering when you `_, subSeg := xray.BeginSubsegment(r.Context(), "http.guaranteedSend")`, I bet it will show an error message since there is not context. Could you confirm about this one?

dlsniper:
Hi I haven't seen the message but I'll replicate it either tonight or tomorrow. But the need for this patch remains, regardless of the error message or not.

luluzhao:
@dlsniper , I understand what you mean. However, we didn't do the nil check for segment before is that if the segment or subsegment cannot be create due to some reasons, it will panic unless customer modifies the `ContextMissingStrategy`. I guess in your case, you modified the `ContextMissingStrategy` to ignore `panic` and then the function returned `nil`. However, we didn't implement `ContextMissingStrategy` in `Close` function so that you received `panic`.

dlsniper:
Sorry for dropping this off. I agree that you haven't done the ` ContextMissingStrategy ` there but, as the PR shows, that is missing from a lot more places.

It really is OK to noop these rather than execute the same strategy only once, when one will begin the segment, and not have to repeat it for all the other operations, in order to speed up the execution when the sement is missing.

luluzhao:
@dlsniper , I still have a concern about your PR since it may cause a memory leak. All these six methods return error type. When there is a nil segment or subsegment, the program should panic or at least give a error message because there is an error in the code.

dlsniper:
Thank you for your reply. I'll have a look at the memory leak and fix it.

However I believe that generating and extra error, or worse, panicking, is not a desirable effect. High performance applications would be impacted by this, which is what this PR attempts to fix. The segment creation using the ` ContextMissingStrategy ` should be enough for users. I don't see how they could be in a position to close a segment without opening it to begin with.

luluzhao:
close it for now since it hasn't been updated more than a year and feel free to reopen it.

PR #40, opened 2018-01-30
Explicitly initialize plugins (fixes #38)

### LINKED ISSUE #38: Do not use init() in ec2 and ecs packages
Adding the following two imports to my application:

```go
	_ "github.com/aws/aws-xray-sdk-go/plugins/ec2"
	_ "github.com/aws/aws-xray-sdk-go/plugins/ecs"
```

will make it take up to 10 seconds to timeout before it can continue loading when I'm working on my local machine and not in an EC2 or ECS instance.

I would like to argue for removing the implicit behaviour those two imports have and manually adding an idempotent ` Init() ` function on each package which will allow users to call those functions whenever needed. It might not make for a "nice" experience but I'll take adding an extra couple of lines of code any day over 10 seconds of waiting time on application boot any single day.

dlsniper:
This allows a user to import plugins on local development environments which do not have access to AWS metadata without stalling the process at boot time.

Each of the three plugins will need to be imported and initialized manually.
The API will safeguard against multiple requests but it's not goroutine safe.

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,18 +53,22 @@ import (
   "github.com/aws/aws-xray-sdk-go/xray"
 
   // Importing the plugins enables collection of AWS resource information at runtime.
-  // Every plugin should be imported after "github.com/aws/aws-xray-sdk-go/xray" library.
-  _ "github.com/aws/aws-xray-sdk-go/plugins/ec2"
-  _ "github.com/aws/aws-xray-sdk-go/plugins/beanstalk"
-  _ "github.com/aws/aws-xray-sdk-go/plugins/ecs"
+  "github.com/aws/aws-xray-sdk-go/plugins/ec2"
+  "github.com/aws/aws-xray-sdk-go/plugins/beanstalk"
+  "github.com/aws/aws-xray-sdk-go/plugins/ecs"
 )
 
 func init() {
-  xray.Configure(xray.Config{
+	// Enable the collection of AWS resource information
+    ec2.Init()
+    beanstalk.Init()
+    ecs.Init()
+    
+    xray.Configure(xray.Config{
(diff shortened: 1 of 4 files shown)


ScottMansfield:
why not surround the init code with sync.Once to make them concurrent safe?

dlsniper:
> why not surround the init code with sync.Once to make them concurrent safe?

That's a very good point, I'll update the PR to reflect this, thank you!

docmerlin:
Yah, this is a better way of doing it.  `init` forces it to be at starttime and can cause conflicts occasionally.

dlsniper:
> Yah, this is a better way of doing it. `init` forces it to be at starttime and can cause conflicts occasionally.

This is not why I wanted to change this, please read #38.

README.md
```diff
@@ -53,18 +53,22 @@ import (
   "github.com/aws/aws-xray-sdk-go/xray"
 
   // Importing the plugins enables collection of AWS resource information at runtime.
-  // Every plugin should be imported after "github.com/aws/aws-xray-sdk-go/xray" library.
-  _ "github.com/aws/aws-xray-sdk-go/plugins/ec2"
-  _ "github.com/aws/aws-xray-sdk-go/plugins/beanstalk"
-  _ "github.com/aws/aws-xray-sdk-go/plugins/ecs"
+  "github.com/aws/aws-xray-sdk-go/plugins/ec2"
```
luluzhao: Could you please rename these three plugin package since directly changing the plugin behavior under the same package name will not result in any compile or runtime errors for consumers, and will silently fail to record AWS metadata on their segments. Like `"github.com/aws/aws-xray-sdk-go/awsplugins/ec2"`

PR #43, opened 2018-02-22
Default config looks at the environment variable for the daemon address

drocamor:
Hi,

This change makes the SDK's default configuration try to get the daemon address from the environment. It should revert to the old default (127.0.0.1:2000) if the environment variable is missing or if it is malformed.

Let me know if this needs some revision before you merge.

Thanks!

### DIFF
--- a/xray/config.go
+++ b/xray/config.go
@@ -36,13 +36,28 @@ type SDK struct {
 var globalCfg = newGlobalConfig()
 
 func newGlobalConfig() *globalConfig {
-	ret := &globalConfig{
-		daemonAddr: &net.UDPAddr{
+
+	// Try to get the X-Ray daemon address from an environment variable
+	var daemonAddress *net.UDPAddr
+	var err error
+
+	environmentDaemonAddress := os.Getenv("AWS_XRAY_DAEMON_ADDRESS")
+
+	// Try to parse the address
+	daemonAddress, err = net.ResolveUDPAddr("udp", environmentDaemonAddress)
+
+	// If the environment variable is empty or we can't parse the address, use 127.0.0.1:2000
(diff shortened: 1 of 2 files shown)


drocamor:
After some thought, I think this should panic if the contents of the environment variable are not a valid address. What do you think?

luluzhao:
@drocamor, thank you so much for reaching out to us regarding your idea. I looked the PR and I am sorry I didn't get the idea why you want SDK to get default daemon address from environment variable first rather than giving a default value? In my opinion, we shouldn't pass customer defined value to a default variable and instead, it should be pre-defined.

xray/config.go
```diff
@@ -37,29 +37,28 @@ var globalCfg = newGlobalConfig()
 
 func newGlobalConfig() *globalConfig {
 
-	// Try to get the X-Ray daemon address from an environment variable
-	var daemonAddress *net.UDPAddr
-	var err error
-
-	environmentDaemonAddress := os.Getenv("AWS_XRAY_DAEMON_ADDRESS")
+	ret := &globalConfig{
```
luluzhao: Here we should call `loadLogConfig("", "")` method. One is to get the default value, the other is to init log with default setting.

xray/config_test.go
```diff
@@ -110,6 +110,15 @@ func TestEnvironmentConfigureParameters(t *testing.T) {
 
 }
 
+func TestInvalidEnvironmentDaemonAddress(t *testing.T) {
+
```
luluzhao: nitpick.... could you remove the empty line at the beginning and end of the method?

drocamor:
OK, I made those changes. Let me know what you think!

PR #44, opened 2018-03-03
Fix race with subsegments

creack:
Solves a race condition when using subsegments in goroutines.

```
WARNING: DATA RACE
Read at 0x00c420256798 by goroutine 26:
vendor/github.com/aws/aws-xray-sdk-go/xray.BeginSubsegment()
vendor/github.com/aws/aws-xray-sdk-go/xray/segment.go:184 +0x347
lib.worker()
lib/health.go:74 +0x296
lib.(*Controller).GetHealthcheckHandler.func2()
lib/health.go:140 +0x80
Previous write at 0x00c420256798 by goroutine 23:
vendor/github.com/aws/aws-xray-sdk-go/xray.BeginSubsegment()
vendor/github.com/aws/aws-xray-sdk-go/xray/segment.go:184 +0x369
```

Running `make test` with `-race` on master:

```
--- FAIL: TestErrorCapture (0.00s)
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
        panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x30 pc=0x15cd21b]

goroutine 51 [running]:
testing.tRunner.func1(0xc420286000)
        /usr/local/opt/go/libexec/src/testing/testing.go:742 +0x567
panic(0x166a660, 0x1a0c480)
        /usr/local/opt/go/libexec/src/runtime/panic.go:505 +0x24a
github.com/aws/aws-xray-sdk-go/xray.TestErrorCapture(0xc420286000)
        /Users/gcharmes/go/src/github.com/aws/aws-xray-sdk-go/xray/capture_test.go:76 +0x42b
testing.tRunner(0xc420286000, 0x1709990)
        /usr/local/opt/go/libexec/src/testing/testing.go:777 +0x16e
created by testing.(*T).Run
        /usr/local/opt/go/libexec/src/testing/testing.go:824 +0x565
FAIL    github.com/aws/aws-xray-sdk-go/xray     0.063s
make: *** [test] Error 1
```

Update: as the build was failing for the PR, I added a .travis.yml, cleaned up the Makefile and fixed the lint/fmt/vet errors.

Update2: There is a race in the tests, not detected by the race tool. After fixing the test to avoid panic, I got (after ~10 runs):

```
$> while go test -failfast -count=1 -cover -race github.com/creack/aws-xray-sdk-go/xray; do; done
[...]
--- FAIL: TestErrorCapture (0.00s)
        Error Trace:    capture_test.go:84
        Error:          Expected value not to be nil.
        Messages:       Missing cause from error subsegment.

--- FAIL: TestPanicCapture (0.00s)
        Error Trace:    capture_test.go:114
        Error:          Not equal: "MyPanic" (expected)
                                != "MyError" (actual)

        Error Trace:    capture_test.go:115
        Error:          Not equal: "panic" (expected)
                                != "error" (actual)

        Error Trace:    capture_test.go:116
        Error:          Not equal: "TestPanicCapture.func1.2" (expected)
                                != "TestErrorCapture.func1" (actual)

        Error Trace:    capture_test.go:118
        Error:          Not equal: "TestPanicCapture.func1" (expected)
                                != "TestErrorCapture" (actual)

        Error Trace:    capture_test.go:119
        Error:          Not equal: "TestPanicCapture" (expected)
                                != "tRunner" (actual)

FAIL
```

It was due to the TestDaemon leaking from one test to another. I did a minor refactor so each test has it's own daemon.


Makefile
```diff
@@ -0,0 +1,48 @@
+GO_VET_CMD      = go tool vet --all -shadow
+GO_DIRS         = $(shell ls -d */ | \grep -v vendor)
+GO_PKGS         = $(shell go list ./... | \grep -v vendor)
+GO_SRCS         = $(shell find . -type f -name '*.go')
+
+# assets removes the existing bindata.go and regenerates it.
+.PHONY          : assets
```
luluzhao: nitpick... There should be no space between `.PHONY` and `: assets`.

Makefile
```diff
@@ -0,0 +1,48 @@
...
+assets          : clean resources/bindata.go
```
luluzhao: nitpick... I suggest you start a new line for following command, like:
assets: clean resources/bindata.go

Makefile
```diff
@@ -0,0 +1,48 @@
...
+
+# test runs the lint/vet/fmt and the tests.
+.PHONY          : test
+test            : vet lint fmt imports
+test            : resources/bindata.go ${GO_SRCS}
```
luluzhao: I suggest do know touch `resources/bindata.go` file since it is generated by another go library and I don't think we should do a test for it.

xray/segment.go
```diff
@@ -227,34 +226,34 @@ func (seg *Segment) Close(err error) {
 	seg.Unlock()
 
 	if err != nil {
-		seg.AddError(err)
+		_ = seg.AddError(err) // Store only.
```
luluzhao: why we add `_=` before calling the method?

xray/lambda.go
```diff
@@ -68,21 +69,19 @@ func createFile(dir string, name string) (string, error) {
...
-		if e != nil {
-			return filePath, e
-		} else {
-			var file, err = os.Create(filePath)
-			if err != nil {
-				return filePath, err
-			}
-			file.Close()
-			return filePath, nil
+		if e1 := os.MkdirAll(dir, os.ModePerm); e1 != nil {
```
luluzhao: Could you provide unit tests for code refactoring about this part?

xray/segment_model.go
```diff
@@ -21,7 +21,7 @@ type Segment struct {
 	sync.Mutex
 	parent           *Segment
 	openSegments     int
-	totalSubSegments int
+	totalSubSegments int64
```
luluzhao: I am wondering why we have to change the variable type at here?

Makefile
creack: I tend to keep the target and deps aligned, but if you prefer, I can remove it.

Makefile
creack: You mean new line between PHONY and assets?

luluzhao:
@creack, thank you so much for sending out this PR. I am sorry I approved your PR by mistake. I think we still have to spend some time working on this PR before merging to our branch. I already submit a review and free feel to add some comments or changes into it.

Makefile
creack: In here, bindata.go is a dependency, i.e. if it changes, then the test will re-run.

xray/segment.go
creack: To comply with errcheck linter and be explicit on the error ignore. It makes it clear it is a vonluntary ignore vs a mistake.

xray/lambda.go
creack: I'll see what I can do, but I suspect it is already tested. this minor refactor doesn't change the logic.

xray/segment_model.go
creack: Changed the type so we can use the atomic package directly and make it thread safe.

Makefile
luluzhao: https://www.gnu.org/software/make/manual/html_node/Phony-Targets.html
I suggest you read this link and format the Makefile. :) I prefer we follow the Makefile convention.

Makefile
luluzhao: https://www.gnu.org/software/make/manual/html_node/Phony-Targets.html
No, I mean start a new line for `clean resources/bindata.go`

Makefile
creack: I am quite familiar with the make docs ;) However, even in that link, they do not have a consistent style, some time using spaces some time not. But no worries, I'll push the change and remove the spaces.

Makefile
creack: I am sorry, I don't understand what you mean. Could you provide an example? You mean a new line between `assets:` and `clean resources/bindata.go`?

Makefile
luluzhao: I see. Could we rename the resources/bindata.go section? Start a `.PHONY` section and rename to some keywords other than `resources/bindata.go`(The section I mentioned is at the end of Makefile)

creack: So you want to remove the `go-bindata` call to generate `resources/bindata.go` and manually generate it instead?

Makefile
luluzhao: You can find example in the link I attached. Like below:
.PHONY: clean
clean:
        rm *.o temp

creack: I really don't understand sorry, the code is already like this.

```
.PHONY: clean
clean:
        rm -f resources/bindata.go
```

But as you requested to remove the bindata generation, I"ll just remove the target altogether.

Makefile
luluzhao: No. I don't want you to remove go-bindata call. I just want the following part:

resources/bindata.go    : $(shell find resources -type f -name '*.json')
		rm -f $@
 		go-bindata -o resources/bindata.go -pkg resources resources/
 	        goimports -w resources/bindata.go
		gofmt -w -s resources/bindata.go

to have same `.PHONY` section like what you have done for `fmt` `clean` etc.

creack: `resources/bindata.go` is a file, not a PHONY target. Why would it be in the phony section? It would remove the whole purpose of the makefile.

xray/lambda.go
luluzhao: Have you test this part already? I think it's more safe if we test the code before we merge. We cannot 100% sure that our change won't break anything even if it looks like there is no logic error.

Makefile
luluzhao: Thanks. :)

xray/lambda.go
creack: The only reason I changed it was to comply with golint and govet. I reverted to the original code and disabled the linters.

creack:
I think all your coment have been addressed. Let me know if I missed something.

Makefile
```diff
@@ -3,14 +3,10 @@ GO_DIRS = $(shell ls -d */ | \grep -v vendor)
 GO_PKGS = $(shell go list ./... | \grep -v vendor)
 GO_SRCS = $(shell find . -type f -name '*.go')
 
-# assets removes the existing bindata.go and regenerates it.
```
luluzhao: Sorry, I think you misunderstand what I said. We need assets generation but I don't want to format the generated file `bindata.go`.

Makefile
```diff
@@ -35,14 +31,3 @@ imports: ${GO_SRCS}
...
-
-.PHONY: clean
-clean:
-	rm -f resources/bindata.go
-
-# generate & format bindata.go.
-resources/bindata.go: $(shell find resources -type f -name '*.json')
-	rm -f $@
-	go-bindata -o resources/bindata.go -pkg resources resources/
-	goimports -w resources/bindata.go
```
luluzhao: I think we don't have to do `goimports` and `gofmt` for the generated `bindata.go` file.

Makefile
creack: I am sorry but I can't do that, if I am to add the generation, it will be properly formatted. However, as it was not there before and it is not there now so I'd say we can move ahead. Feel free to add it in a later time if you'd like. I am mostly concerned about the race condition.

creack: Sorry about that because of case change, I missed it. Makefile reverted.

creack: However, note that 1) the case is wrong, 2) make vet fails, 3) make lint fails

luluzhao: Unit tests failed to pass. Have you run unit tests on your local machine? What is the result?

creack: Work fine. However, as the build was failing before too.

luluzhao: Let's fix this issue before we merge since this is the first time I saw the tests are failing. The reason it's not passing before is that we didn't finish the implementation about travis. However, you finished the rest part of it and it should work. Something goes wrong. I will check in your commits in my local box to see what is going on.

creack: The tests are failing on master... just run them in a loop for a few seconds and you will see them fail.

On master:

```
while go test -count=1 ./xray; do; done
```

Or make them fail faster:

```
while go test -race -count=1 ./xray; do; done
```

PR #56, opened 2018-05-04
Add timing subsegment for request headers to tracer

rengers:
* Includes the WroteHeaders() event
* This event fires after the Transport has written the request headers
* I have included it as a subsegment of the request subsegment

### DIFF
--- a/xray/httptrace_go17.go
+++ b/xray/httptrace_go17.go
@@ -18,13 +18,14 @@ import (
 
 // HTTPSubsegments is a set of context in different HTTP operation.
 type HTTPSubsegments struct {
-	opCtx       context.Context
-	connCtx     context.Context
-	dnsCtx      context.Context
-	connectCtx  context.Context
-	tlsCtx      context.Context
-	reqCtx      context.Context
-	responseCtx context.Context
+	opCtx         context.Context
+	connCtx       context.Context
+	dnsCtx        context.Context
+	connectCtx    context.Context
+	tlsCtx        context.Context
+	reqCtx        context.Context
+	reqHeadersCtx context.Context
+	responseCtx   context.Context
 }
 
 // NewHTTPSubsegments creates a new HTTPSubsegments to use in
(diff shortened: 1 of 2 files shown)


luluzhao:
@rengers , thank you for submitting this PR and we appreciate your effort. Could you tell us more about the idea of this PR and how would this change help your team to diagnose issues. What's more, would you mind to make this change configurable?

rengers:
Hi @luluzhao ,

Thanks for taking a look.

The golang net/http exposes this data, so I thought it would make sense to include in the net http trace, seeing as we are already including all the other timing metrics.

In terms of the data itself, the wrote-header time shows how much time is being spent putting the headers on the wire. It should be very small. A large value could indicate an issue with large headers or failing to write to the connection itself.

Additionally if a failure occurs while writing, knowing if it occurs before or after writing the headers may be useful for debugging.

Source:
https://golang.org/pkg/net/http/httptrace/#ClientTrace

luluzhao:
Hello @rengers , sorry to get back later to you. I was able to test this PR but here I found out the new `request-headers` subsegment had the same duration with its parent `request` subsegment. Since you mentioned above it should be very small. Does that mean in most of the no error case, these two values are always the same?

rengers:
Hey @luluzhao sorry I didn't see the notification.

Yes I think under normal circumstances the `WroteHeaders` time should be really small. 

From the docs:
```
// WroteHeaders is called after the Transport has written
// all request headers.
```
```
// WroteRequest is called with the result of writing the
// request and any body. It may be called multiple times
// in the case of retried requests.
```

xray/httptrace_go17.go
```diff
@@ -112,11 +113,20 @@ func (xt *HTTPSubsegments) GotConn(info *httptrace.GotConnInfo, err error) {
...
+			xt.reqHeadersCtx, _ = BeginSubsegment(xt.reqCtx, "request-headers")
 		}
 
 	}
 }
 
+// WroteHeaders closes the request-headers subsegment if the HTTP operation
+// subsegment is still in progress.
+func (xt *HTTPSubsegments) WroteHeaders() {
+	if xt.reqHeadersCtx != nil && GetSegment(xt.opCtx).InProgress {
```
luluzhao: Could we call [safeInProgress](https://github.com/aws/aws-xray-sdk-go/blob/master/xray/segment.go#L337) instead of directly access `InProgress` since it will cause unclose `in_progress` subsegment issue.

rengers: Updated to use safeInProgress

luluzhao:
@rengers thank you for your quick update. I will follow up with our PM regarding this issue and get back to you soon. Thank you for your patience in the meantime.

ss2305:
@rengers Thanks for the PR.  Our GO SDK currently generates quite a bit of information and we try to keep the segment light to alleviate performance bottlenecks .
We don't see too many use cases would require something as intricate as the wroteheader time. Can you shine some light on your specific use case?
Also looking at the code it seems you are trying to generate a new subsegment ? Correct me if I am wrong.

rengers:
Hi @ss2305,

You can read what our use case for this data was above in the thread.
https://github.com/aws/aws-xray-sdk-go/pull/56#issuecomment-389351140

If you do not intend to merge this PR, please close it.

Thanks,
Ross

luluzhao:
Hi @rengers Our Go segment is already pretty heavy and given the nature of how the traces that get passed around need to be lightweight we will have to close this for now. Thanks for your engagement and feel free to reach out .

Issue #50, opened 2018-04-08
"segment cannot be found" in lambda function


aleist:
I'm new to X-Ray and am trying to integrate it into a Lambda function called from API Gateway as a Lambda proxy integration request. My code follows this example:

```go
xray.Configure(xray.Config{LogLevel: "trace"})
sess := session.Must(session.NewSession())
dynamo := dynamodb.New(sess)
xray.AWS(dynamo.Client)
````

This results in the following error being logged `Error Suppressing AWS X-Ray context missing panic: failed to begin subsegment named 'dynamodb': segment cannot be found.`, which is followed by a panic `runtime error: invalid memory address or nil pointer dereference` in the xray SDK when the first DynamoDB operation executes.

I can "fix" this by creating a new segment with `xray.BeginSegment`, which puts a new service node in the graph.

I have tried this both with "active tracing" enabled and disabled in the Lambda config. When enabled, I get 2 more nodes for the Lambda service and function in the graph, but they're not connected to the new segment I create in the function as described above.

I am not sure if I'm doing something wrong (in which case the example would seem wrong too) or if this is a bug.

It also worries me that this results in a panic that propagates to my Lambda, because even if I fix this particular problem, I don't like the idea of a library that just instruments my code potentially causing a fatal error of the lambda.

luluzhao:
Hi @aleist, thank you for contacting us regarding your issue and I am sorry for the inconvenience. When using X-Ray SDK inside a lambda function, lambda will create a segment and X-Ray SDK will create corresponding subsegment for customer's application, which means you don't have to call `xray.BeginSegment` inside a lambda function. Here I attach an example how to use X-Ray SDK inside a lambda function.

```go
func HandleRequest(ctx context.Context, name string) (string, error) {
    xray.Configure(xray.Config{LogLevel: "trace",})
    sess := session.Must(session.NewSession())
    dynamo := dynamodb.New(sess)
    xray.AWS(dynamo.Client)
    input := &dynamodb.PutItemInput{
        Item: map[string]*dynamodb.AttributeValue{
            "a": {
                S: aws.String("xray"),
            },
        },
        TableName: aws.String("xray"),
    }

    _, err := dynamo.PutItemWithContext(ctx, input)
    if err != nil {
        return name, err
    }

    return fmt.Sprintf("Hello %s!", name), nil
}
````
[...]

PR #59, opened 2018-06-23
Update readme to reflect Lambda supporting version

artemnikitin:
In https://github.com/aws/aws-xray-sdk-go/issues/50 said what AWS Lambda support is avaialble starting from v1.0.0-rc.1 and higher. If you are using `dep`, then by default it will select version 0.9.4 without Lambda support.

This change will help people to properly start using SDK for Lambda

### DIFF
--- a/README.md
+++ b/README.md
@@ -147,6 +147,10 @@ func main() {
 
 **Lambda**
 
+```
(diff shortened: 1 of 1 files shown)


luluzhao:
@artemnikitin, I merged this change and thank you for your contribution.

Issue #71, opened 2018-10-17
Data race when use http client


domashevskiy:
Reproduced on:
- SDK version: V1.0.0-rc.8.
- GO version: go1.11.1 linux/amd64

When build with - race flag, data race warning occurs upon sending http request:

WARNING: DATA RACE
Read at 0x00c000a201b0 by goroutine 56:
  vendor/github.com/aws/aws-xray-sdk-go/xray.(*HTTPSubsegments).GotFirstResponseByte()
      vendor/github.com/aws/aws-xray-sdk-go/xray/httptrace.go:159 +0x47
  privacyapi/vendor/github.com/aws/aws-xray-sdk-go/xray.NewClientTrace.func10()
      vendor/github.com/aws/aws-xray-sdk-go/xray/httptrace.go:212 +0x41
  net/http.http2traceFirstResponseByte()
      /usr/local/go/src/net/http/h2_bundle.go:3035 +0x7d
  net/http.(*http2clientConnReadLoop).processHeaders()
      /usr/local/go/src/net/http/h2_bundle.go:8360 +0x52a
  net/http.(*http2clientConnReadLoop).run()
      /usr/local/go/src/net/http/h2_bundle.go:8292 +0x87d
  net/http.(*http2ClientConn).readLoop()
      /usr/local/go/src/net/http/h2_bundle.go:8189 +0x8c

Previous write at 0x00c000a201b0 by main goroutine:
  vendor/github.com/aws/aws-xray-sdk-go/xray.(*HTTPSubsegments).WroteRequest()
      vendor/github.com/aws/aws-xray-sdk-go/xray/httptrace.go:152 +0x156
 vendor/github.com/aws/aws-xray-sdk-go/xray.NewClientTrace.func9()
      vendor/github.com/aws/aws-xray-sdk-go/xray/httptrace.go:209 +0x55
  net/http.http2traceWroteRequest()

luluzhao:
@domashevskiy , sorry for the late response. I merged @t-yuki PR and will make sure bring this fix to next release.

PR #72, opened 2018-10-18
fix datarace for http2

t-yuki:
*Issue #71

*Description of changes: Unit test and quick fix of http2 datarace reported in #71
However, it may have other problems by race since this is only a trivial patch.
Also, since httptest.NewRequest is for server side incoming requests, as noted in the document, we should use http.NewRequest.

### DIFF
--- a/xray/client_test.go
+++ b/xray/client_test.go
@@ -10,6 +10,8 @@ package xray
 
 import (
 	"context"
+	"crypto/tls"
+	"crypto/x509"
 	"encoding/json"
 	"fmt"
 	"io/ioutil"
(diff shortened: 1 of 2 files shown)


xray/client_test.go
```diff
@@ -225,3 +228,42 @@ func TestRoundTrip_reuse_datarace(t *testing.T) {
 	}
 	wg.Wait()
 }
+
+func TestRoundTrip_http2_datarace(t *testing.T) {
```
luluzhao: Nitpick.... Could we rename this function to be "TestRoundTripHttp2Datarace"? Since we haven't used underscore in naming our functions, it would be good to keep consistency.

PR #76, opened 2018-12-20
Makes it possible to pass customized emitter

haotianw465:
See title


xray/default_emitter.go
```diff
@@ -0,0 +1,103 @@
...
+	initLambda()
+	d := &DefaultEmitter{}
+	d.RefreshEmitterWithAddress(raddr)
+	return d, nil
+}
+
+// RefreshEmitterWithAddress dials UDP based on the input UPD address.
+func (de *DefaultEmitter) RefreshEmitterWithAddress(raddr *net.UDPAddr) {
+	de.Lock()
+	de.conn, _ = net.DialUDP("udp", nil, raddr)
```
luluzhao: Could we don't ignore the error when calling the DialUDP method?

PR #81, opened 2019-01-25
Explicit plugins

### LINKED ISSUE #38: Do not use init() in ec2 and ecs packages

muirrn:
Fixes #38, supersedes #40.

Add explicit versions of the aws metadata plugins so users can conditionally load the plugins in prod/dev.

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -4,5 +4,6 @@ The following people have contributed to the AWS X-Ray SDK for Go's design and/o
 * Bilal Khan
 * James Bowman
 * Lulu Zhao
+* Muir Manders
 * Raymond Lin
 * Rohit Banga
--- a/README.md
+++ b/README.md
@@ -20,7 +20,7 @@ If you also want to install SDK's testing dependencies. They can be installed us
 go get -u -t github.com/aws/aws-xray-sdk-go/...
 ```
 
-You may also use [Glide](https://github.com/Masterminds/glide) to manage dependencies by using 
+You may also use [Glide](https://github.com/Masterminds/glide) to manage dependencies by using
 
 ```
 glide install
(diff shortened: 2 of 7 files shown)


luluzhao:
@muirrn , thank you so much for sending out PR to us. Feel free to send more PRs in future and we will prioritize them. Since the explicit loading plugins is an breaking change so we want to merge it once we are at GA.

muirrn:
Thank you for responding.

This is not a breaking change because it maintains the same behavior for the existing plugins. The explicit plugins are located under a new path /awsplugins instead of /plugins (as suggested by https://github.com/aws/aws-xray-sdk-go/issues/38#issuecomment-367092739).

luluzhao:
Hi @muirrn , the reason I said it is a breaking change since if customer update the version of X-Ray Go SDK, the old way(automatically start adding information) will not work. The explicit way need customer to modify their application and call `Init()` method to start tracking information.

muirrn:
There is no change in behavior for existing users. The existing plugins are still initialized at package load automatically.

If the user elects to import the **new** plugins under /awsplugins then they have to explicitly call `Load()`.

muirrn:
I added a note in the README.md about the old and new plugins.

PR #82, opened 2019-01-31
Logging interface

### LINKED ISSUE #15: logging library
The logging library used in this SDK is pretty terrible.  I would to use a custom logger or at least have something that isn't using inline XML configuration, its 2017.

muirrn:
Fixes #15.

Replace seelog with a custom logger interface. This change is backwards compatible except that the existing LogLevel and LogFormat config fields no longer have any effect. Users will have to reset their min log level if they weren't using the default of "info".

### DIFF
--- a/README.md
+++ b/README.md
@@ -204,6 +204,31 @@ func init() {
 }
 ```
 
+**Logger**
+
+xray uses an interface for its logger:
+
+```go
+type Logger interface {
+  Log(level LogLevel, msg fmt.Stringer)
+}
+
+const (
+  LogLevelDebug LogLevel = iota + 1
+  LogLevelInfo
+  LogLevelWarn
+  LogLevelError
+)
+```
+
+The default logger logs to stdout at "info" and above. To change the logger, call `xray.SetLogger(myLogger)`. There is a default logger implementation that writes to an `io.Writer` from a specified minimum log level. For example, to log to stderr at "error" and above:
+
+```go
+xray.SetLogger(xraylog.NewDefaultLogger(os.Stderr, xraylog.LogLevelError))
+```
+
+Note that the `xray.Config{}` fields `LogLevel` and `LogFormat` are deprecated and no longer have any effect.
+
 ## License
 
 The AWS X-Ray SDK for Go is licensed under the Apache 2.0 License. See LICENSE and NOTICE.txt for more information.
--- a/awsplugins/beanstalk/beanstalk.go
+++ b/awsplugins/beanstalk/beanstalk.go
@@ -12,8 +12,8 @@ import (
 	"encoding/json"
 	"io/ioutil"
 
+	"github.com/aws/aws-xray-sdk-go/internal/logger"
 	"github.com/aws/aws-xray-sdk-go/internal/plugins"
-	log "github.com/cihub/seelog"
 )
 
 const Origin = "AWS::ElasticBeanstalk::Environment"
@@ -29,14 +29,14 @@ func addPluginMetadata(pluginmd *plugins.PluginMetadata) {
 
 	rawConfig, err := ioutil.ReadFile(ebConfigPath)
 	if err != nil {
-		log.Errorf("Unable to read Elastic Beanstalk configuration file %s: %v", ebConfigPath, err)
+		logger.Errorf("Unable to read Elastic Beanstalk configuration file %s: %v", ebConfigPath, err)
 		return
 	}
 
 	config := &plugins.BeanstalkMetadata{}
 	err = json.Unmarshal(rawConfig, config)
 	if err != nil {
-		log.Errorf("Unable to unmarshal Elastic Beanstalk configuration file %s: %v", ebConfigPath, err)
+		logger.Errorf("Unable to unmarshal Elastic Beanstalk configuration file %s: %v", ebConfigPath, err)
 		return
 	}
 
--- a/internal/logger/logger.go
+++ b/internal/logger/logger.go
@@ -1,0 +1,78 @@
+// Copyright 2017-2017 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License"). You may not use this file except in compliance with the License. A copy of the License is located at
+//
+//     http://aws.amazon.com/apache2.0/
+//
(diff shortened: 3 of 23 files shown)


muirrn:
Hi - can someone work with me on this? I think it resolves all issues I know about with the logger.

luluzhao:
Hi @muirrn, I am looking at it right now and will you let you about my feedback ASAP.

xraylog/xray_log.go
```diff
@@ -0,0 +1,77 @@
...
+	case LogLevelDebug:
+		return "DEBUG"
+	case LogLevelInfo:
+		return "INFO"
+	case LogLevelWarn:
+		return "WARN"
+	case LogLevelError:
+		return "ERROR"
+	default:
+		return fmt.Sprintf("UNKNOWN<%d>", ll)
```
luluzhao: I would suggest by default, the local level should be `INFO` even if the format cannot be recognized.

awsplugins/beanstalk/beanstalk.go
```diff
@@ -12,8 +12,8 @@ import (
 	"encoding/json"
 	"io/ioutil"
 
+	"github.com/aws/aws-xray-sdk-go/internal/logger"
```
luluzhao: Could we directly have a folder named `logger` under `aws-xray-sdk-go` folder instead of being a sub folder inside `internal`?

internal/logger/logger.go
```diff
@@ -0,0 +1,78 @@
...
+//
+// or in the "license" file accompanying this file. This file is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the specific language governing permissions and limitations under the License.
+
+package logger
+
+import (
+	"fmt"
+	"os"
+
+	"github.com/aws/aws-xray-sdk-go/xraylog"
```
luluzhao: I would suggest you keep all logger related file into one folder instead of split into two folders.

README.md
```diff
@@ -204,6 +204,31 @@ func init() {
...
+  LogLevelInfo
+  LogLevelWarn
+  LogLevelError
+)
+```
+
+The default logger logs to stdout at "info" and above. To change the logger, call `xray.SetLogger(myLogger)`. There is a default logger implementation that writes to an `io.Writer` from a specified minimum log level. For example, to log to stderr at "error" and above:
+
+```go
+xray.SetLogger(xraylog.NewDefaultLogger(os.Stderr, xraylog.LogLevelError))
```
luluzhao: I would suggest you could include the logger setting into `xray.Config{}` to keep consistency. It can be inconvenient for customers to configure setting through different way.

xraylog/xray_log.go
muirrn: This is just the `String()` of the format, so the return value will not affect what log messages show up. The default log level is already "info" (this method is just for displaying the log level as a string).

It shouldn't be possible in normal circumstances to get an unknown log level. But if it does happen somehow, something weird is happening and logging may not be working as expected,  so this tells the developer/user that something is wrong.

awsplugins/beanstalk/beanstalk.go
muirrn: I made it internal so users of xray can not access the internal logging functions (we don't want packages outside xray to be able to use our logger).

internal/logger/logger.go
muirrn: I split it in two so we can have the public API for implementing the logger, and the private API for writing to the logger.

README.md
muirrn: I wanted to keep it in the config, but I decided it was not safe. The logger is something you only want to set up once, but the config can be changed many times. In fact, you can set the config in the context using `ContextWithConfig`, so you can have multiple segments at the same time using different configs. Since the logger is global, you can't have two different logging configs at the same time. Therefore, to make sure it is safe and avoid race conditions, the logger must be set up one time only in `init()` method.

yogiraj07:
Hi @muirrn , I am currently working on this PR. Will post an update soon.

README.md
yogiraj07: Hi @muirrn ,
I would still prefer having logger setting into  `xray.Config{}`. This adds flexibility for users if they want logger per context or at global level. This also eliminates new method to be introduced for configuring logger apart from current way of  `xray.Config{}` and helps in achieving both, per context as well as global level logging functionality. For example [`BeginSegment`](https://github.com/aws/aws-xray-sdk-go/blob/9779f16ea01fdfc11c54971ed8547dbaa442bb42/xray/segment.go#L59) accepts context struct, and if present, configuration is applied from that context else from global config. Could you explain more on safety and race conditions issues with this approach, since logger instance would be created/accessed per context?

yogiraj07: Hi @muirrn ,
Thank you for the explanation. I just wanted to make sure that we think through all scenarios. Your observation is correct about seelog usage.

xraylog/xray_log.go
yogiraj07: Will `UNKNOWNLOGLEVEL` be more informative?

xraylog/xray_log.go
```diff
@@ -0,0 +1,77 @@
...
+	w        io.Writer
+	minLevel LogLevel
+}
+
+func (l *defaultLogger) Log(ll LogLevel, msg fmt.Stringer) {
+	if ll < l.minLevel {
+		return
+	}
+
+	fmt.Fprintf(l.w, "%s [%s] %s\n", time.Now().Format(time.RFC3339), ll, msg)
```
yogiraj07: Should this be thread safe ? 

In `log` package,  [Output()](https://golang.org/src/log/log.go?s=8948:8992#L284) which actually writes to writer is thread safe. It seems [`FprintF`](https://github.com/golang/go/blob/master/src/fmt/print.go#L197) is not thread safe.

xraylog/xray_log.go
muirrn: Done.

xraylog/xray_log.go
muirrn: Done.

yogiraj07:
Hi @muirrn ,
Thank you for working with me. I have merged the PR.

Best,
Yogi

PR #83, opened 2019-02-11
Don't try to udp dial emitter at package load time

### LINKED ISSUE #79: Remove log during package import.
xray/default_emitter.go

func (de *DefaultEmitter) RefreshEmitterWithAddress(raddr *net.UDPAddr) {
	de.Lock()
	de.conn, _ = net.DialUDP("udp", nil, raddr)
	log.Infof("Emitter using address: %v", raddr) . <<<<<<< 
	de.Unlock()
}

Can we please get rid of the log line above?
That particular log shows up as part of package import and it is really annoying, for example if you build an app to print out json payloads for another script to consume it.....  that line will always add a line .. "Emitter using address". And because it is done as part of package import it is hard to disable it.

muirrn:
This is the last non-trivial side effect that happens at package load
time (I think). The problem is the user has not had a chance to
Configure() xray yet, so it always tries dialing the default emitter
address (and logs to stdout before user has configured logger).

Instead of trying to dial a default emitter UDP address at package
load time, defer that operation until we try to Emit() our first
message. We also log and return on errors returned by DialUDP().

Fixes #79

### DIFF
--- a/xray/default_emitter.go
+++ b/xray/default_emitter.go
@@ -72,6 +72,7 @@ func (de *DefaultEmitter) Emit(seg *Segment) {
 
 		if de.conn == nil {
 			if err := de.refresh(de.addr); err != nil {
+				de.Unlock()
 				return
 			}
 		}
--- a/xray/default_emitter.go
+++ b/xray/default_emitter.go
@@ -24,23 +24,35 @@ var Header = []byte(`{"format": "json", "version": 1}` + "\n")
 type DefaultEmitter struct {
 	sync.Mutex
 	conn *net.UDPConn
+	addr *net.UDPAddr
 }
 
 // NewDefaultEmitter initializes and returns a
 // pointer to an instance of DefaultEmitter.
 func NewDefaultEmitter(raddr *net.UDPAddr) (*DefaultEmitter, error) {
 	initLambda()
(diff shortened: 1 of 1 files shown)


xray/default_emitter.go
```diff
@@ -57,6 +69,13 @@ func (de *DefaultEmitter) Emit(seg *Segment) {
 			return b.String()
 		})
 		de.Lock()
+
+		if de.conn == nil {
```
yogiraj07: Should we do this check before `packSegments` call. In high intensive subsegment streaming, the current check will be performed for each batch (though only first check will dial up), or if dial up throws error, we dont parse the tree : small optimization. 
Also in case of error 

```
if err := de.refresh(de.addr); err != nil {
				return
			}
```
I think we should first release the lock before returning, so that subsequent operations acquire the lock.

muirrn: I fixed the missing `Unlock`.

As for the order. I think it is better to put it inside the loop so we don't have to lock and unlock the mutex an additional time prior to the loop. That will improve performance somewhat for the "normal" case, which I think makes more sense than optimizing performance for an error case.

yogiraj07: Sure, fair point.

PR #88, opened 2019-03-11
Setup travis

chrisradek:
*Description of changes:*
This change fixes some issues that were preventing travis builds to run. Ran glide update on package to pull in the http2 test dependencies. Before this change, we had errors due to missing golang.org/x/net http2 packages.

I've removed 'tip' from the travis matrix for now, since some tests are failing on go1.12. Fixes for those will come in a separate PR.


chrisradek:
I changed the config so 1.7 and 1.8 build failures won't fail travis. This will just unblock us so we can start having travis run again while we decide what to do about those versions.

Issue #73, opened 2018-11-01
Replace glide


emirb:
As the [documentation mentions](https://github.com/Masterminds/glide#golang-dep), using Glide for dependency management is not recommended anymore as `dep` is available for some time now. 

I would propose moving to `dep` or even better to Go modules (previously known as `vgo`) which are available since 1.11 release.

BHunter2889:
I would like to add a subtask to this:

- [ ]  fix versioning to meet `dep` standards by removing  `-rc` tagging practice which prevents `dep` from getting latest version.

emirb:
@luluzhao Are you going to skip `dep` support and skip to Go modules directly?

luluzhao:
Hello @emirb , after reading [Go package management tools wiki](https://github.com/golang/go/wiki/PackageManagementTools), I found out Go111Module became the "official experiment" which was "dep tool" before. Thanks for pointing out this and we are definitely going to mark Go modules as priority to solve glide issue. Please stay in tuned.

chrisradek:
The SDK now supports `dep` and go modules as of [V1.0.0-rc.11](https://github.com/aws/aws-xray-sdk-go/releases/tag/V1.0.0-rc.11).

Removing the `-rc` tag will be a separate task once the SDK is no longer in a release candidate stage.

PR #90, opened 2019-03-13
feat: add support for go dep and go modules

chrisradek:
*Issue #, if available:*
#73 

*Description of changes:*
This change adds support for go dep and go modules. It also updates travis to use `dep` in go versions 1.9 and 1.10, while using go modules in go versions 1.11 and 1.12 (and tip).

I did not remove support for `glide` yet, and it's still used with go versions 1.7 and 1.8 in travis.


chrisradek:
Good call-out, I'll update the readme.

PR #97, opened 2019-04-01
Add xray.AWSSession to install handlers on session

logan:
An application has to call xray.AWS for each AWS client it constructs.
This creates opportunities for blind spots if someone forgets to
configure a new client.

The xray.AWSSession installs the same handlers at the Session level.
Clients inherit handlers from the session they're created with. As long
as the application systematically reuses the same session to create
clients, it only needs to install X-Ray handlers once.

### DIFF
--- a/xray/aws.go
+++ b/xray/aws.go
@@ -21,6 +21,7 @@ import (
 
 	"github.com/aws/aws-sdk-go/aws/client"
 	"github.com/aws/aws-sdk-go/aws/request"
+	"github.com/aws/aws-sdk-go/aws/session"
 	"github.com/aws/aws-xray-sdk-go/internal/logger"
 	"github.com/aws/aws-xray-sdk-go/resources"
 )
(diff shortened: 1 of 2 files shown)


luluzhao:
Hey @logan , sorry for the delay. I will review it recently.

xray/aws.go
```diff
@@ -150,33 +151,44 @@ var xRayAfterRetryHandler = request.NamedHandler{
...
 		panic("Please initialize the provided AWS client before passing to the AWSWithWhitelist() method.")
 	}
-	pushHandlers(c)
-	c.Handlers.Complete.PushFrontNamed(xrayCompleteHandler(filename))
+	pushHandlers(&c.Handlers, filename)
+}
+
+// AWSSession adds X-Ray tracing to an AWS session. Clients created under this
+// session will inherit X-Ray tracing.
+func AWSSession(s *session.Session) {
```
luluzhao: Could you add some units to the new methods(AWSSession and AWSSessionWithWhitelist)

awssandra:
Hey Logan, we'd be happy to merge this in with a few edits. Or, we can pick it up. Thanks!

logan:
I've added tests for the two new functions, plus the existing `AWSWithWhitelist`. The whitelist variants do not test whitelist behavior, they just pass in the path to the default whitelist.

Travis doesn't seem to be enqueueing the test runs for this PR. but tests pass locally for me in go 1.12.

luluzhao:
@logan Thank you so much for your contribution. Merged it.

PR #113, opened 2019-05-31
Update lock abstraction of defaultLogger.

cep21:
Just a first small change about how locks are referenced.  This moves
the lock from an extension like parameter of defaultLogger to a
private member

### DIFF
--- a/xraylog/xray_log.go
+++ b/xraylog/xray_log.go
@@ -61,7 +61,7 @@ func NewDefaultLogger(w io.Writer, minLogLevel LogLevel) Logger {
 }
 
 type defaultLogger struct {
-	sync.Mutex
+	mu sync.Mutex
 	w        io.Writer
 	minLevel LogLevel
 }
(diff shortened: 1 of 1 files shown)


cep21:
It appears the master build of aws-xray-sdk-go is failing due to race conditions.  Those are independent of this diff.  We should look into resolving those, but maybe not at the same time as this change.

luluzhao:
@cep21 thank you for submitting this PR. I think we have to fix the data race condition before accepting PRs since once the travis ci is broken, it will prevent others to merge the PR. What's more, by reading the error, it is the first time for us to see it. I think we have to fix the mutex in centralized sampling first.

cep21:
These jenkins failures may be problems with jenkins.  Is it possible to retrigger the build? https://travis-ci.org/aws/aws-xray-sdk-go/jobs/539947195

luluzhao:
@cep21 , I restart the build and from the log I saw it was test time out. Let's see how it goes this time.

PR #115, opened 2019-06-01
Bugfix

cep21:
Address two bugs that are causing travis to fail


cep21:
A more in depth explanation of each bug is attached to the git commit message of each (they didn't get sent to the PR text on github above)

luluzhao:
@cep21 thanks for working on it. I merged it.

PR #117, opened 2019-06-04
Lints

cep21:
Fix a few lints that are somewhat easier to think about.

*Issue #, if available:*

*Description of changes:*


xray/lambda_test.go
```diff
@@ -2,8 +2,9 @@ package xray
 
 import (
 	"context"
-	"github.com/stretchr/testify/assert"
 	"testing"
+
+	"github.com/stretchr/testify/assert"
```
cep21: These changes come from https://godoc.org/golang.org/x/tools/cmd/goimports

cep21:
I think there's a bug in the code for SQL causing this test to fail.

```
		if strings.Contains(u.Path, "@") {
			u, _ = url.Parse(fmt.Sprintf("%s//%s%%2F%s", u.Scheme, u.Host, u.Path[1:]))
		}

		// Strip password from user:password pair in address
		if u.User != nil {
```

The error return is ignored, which is resetting u to `nil` and causing the next `u.User` to nil panic.

luluzhao:
@cep21 you are right. We should catch the error.

luluzhao:
@cep21 , I am wondering why we close this PR?

cep21:
Oh that was a mistake.  I deleted the branch on my end and I think github closed the PR :/ . I reopened the PR

PR #127, opened 2019-06-17
Lambda BeginFacadeSegment in readme

### LINKED ISSUE #126: Lamda handler Context does not contain Segment
### Bug

When an APIGateway enabled lambda function with tracing information passes it's context to `xray.TraceID()` it does not return the trace id from the header. This leads to downstream issues for tracing since `GetSegment` is always nil. 

### Expected Result
`xray.TraceID` should return the same trace ID that's present in the header. 

### Reproduce
tried on `v1.0.0rc-12` 

The following handler illustrates the problem, the output should return the same trace ids. 

```go
package xray

import (
	"context"
	"encoding/json"
	"github.com/aws/aws-lambda-go/events"
)

// Handler creates a new Peer and/or adds a new connectionId to it
func Handler(ctx context.Context, req events.APIGatewayProxyRequest) (events.APIGatewayProxyResponse, error) {

	headerTraceID := req.Headers["X-Amzn-Trace-Id"]
	xrayTraceID := xray.TraceID(ctx)

	out := map[string]string{
		"header":        headerTraceID,
		"lambdaContext": xrayTraceID,
	}

	code := 200
	var body string
	if _body, err := json.Marshal(out); err == nil {
		body = string(_body)
	} else {
		body = err.Error()
		code = 400
	}

	return events.APIGatewayProxyResponse{
		Body:       body,
		StatusCode: code,
	}, nil
}
```

underscorenygren:
fixes #126 

*Description of changes:*
Added a section to the lambda detailing use of `BeginFacadeSegment` and `xray.SQL`

### DIFF
--- a/README.md
+++ b/README.md
@@ -199,6 +199,61 @@ func HandleRequest(ctx context.Context, name string) (string, error) {
 }
 ```
 
+To use the xray calls directly with lambda - for use with sql for example - you must begin a new facade segment and use the resulting context in relevant context-enabled calls.
+
+```go
+package yourpkg
+
+import (
+    "context"
+    "fmt"
+    "github.com/aws/aws-xray-sdk-go/header"
+    "github.com/aws/aws-xray-sdk-go/xray"
+    "time"
+)
+
+func HandleRequest(ctx context.Context, name string) (string, error) {
+
+    facadeSegmentName := "facade"
+    traceIDtr, isTraced := ctx.Value("x-amzn-trace-id").(string)
+
+    msg := "Request not traced"
+
+    if isTraced {
+        msg = fmt.Sprintf("request traced with %s", traceIDtr)
+
+        header := header.FromString(traceIDtr)
+        newCtx, seg := xray.BeginFacadeSegment(ctx, facadeSegmentName, header)
+        if seg == nil {
+            return "error", fmt.Errorf("couldn't make segment")
+        }
+        defer seg.Close(nil)
+
+        db, err := xray.SQL("postgres", "postgres://user:password@host:port/db")
+        if err != nil {
+            return "error", err
+        }
+        defer db.Close()
+
+        id := 123
+        var username string
+        var created time.Time
+        err = db.QueryRow(newCtx, "SELECT username, created_at FROM users WHERE id=?", id).Scan(&username, &created)
+        if err != nil {
+            return "error", err
+        }
+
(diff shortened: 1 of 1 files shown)


luluzhao:
@underscorenygren , thanks for your PR. I have one question. Why you try to call `BeginFacadeSegment` in your application? The method should be called automatically by aws xray go sdk when you doing lambda instrumentation. Could you tell me more about your use case?

README.md
```diff
@@ -199,6 +199,61 @@ func HandleRequest(ctx context.Context, name string) (string, error) {
...
+    facadeSegmentName := "facade"
+    traceIDtr, isTraced := ctx.Value("x-amzn-trace-id").(string)
+
+    msg := "Request not traced"
+
+    if isTraced {
+        msg = fmt.Sprintf("request traced with %s", traceIDtr)
+
+        header := header.FromString(traceIDtr)
+        newCtx, seg := xray.BeginFacadeSegment(ctx, facadeSegmentName, header)
```
ss2305: @underscorenygren You shouldn't need to call the BeginFacadeSegment as this is handled automatically by the go SDK.  The xray.SQL that you added, will ensure that the subsegments the db/sql calls are traced. Can you resubmit with the changes?

underscorenygren:
I tested this in the issue I submitted and that did not work, the
BeginFacade was necessary to make it work. In that case I didn't query SQL,
but figured that was a more complete example. I can simplify this to my
previous example, which called the `xray.IsTracingEnabled` function (might
be called something slightly different, quoting from memory here).

Erik Nygren
(650) 796 6434

README.md
```diff
@@ -199,6 +199,60 @@ func HandleRequest(ctx context.Context, name string) (string, error) {
...
+
+        db, err := xray.SQL("postgres", "postgres://user:password@host:port/db")
+        if err != nil {
+            return "error", err
+        }
+        defer db.Close()
+
+        var username string
+        var created time.Time
+        err = db.QueryRow(newCtx, "SELECT username, created_at FROM users WHERE id=?", id).Scan(&username, &created)
```
luluzhao: at here, you don't need to inject `newCtx`. Just use the `ctx` from `HandleRequest(ctx context.Context, id int)` and it will start generate subsegment.

luluzhao:
@underscorenygren closed this PR for now and you can find the lambda use case in our updated README.

PR #128, opened 2019-06-27
Updating README about Lambda use case

luluzhao:
*Issue #, if available:* https://github.com/aws/aws-xray-sdk-go/issues/126

*Description of changes:*
Updating README to better document Lambda use case


README.md
```diff
@@ -170,8 +218,9 @@ func main() {
 For Lambda support use version v1.0.0-rc.1 and higher
 ```
 
-Use AWS X-Ray SDK for Go to generate custom subsegments inside AWS Lambda handler function.
+Regarding Lambda integration, lambda will be responsible for generating segments for customers and send them. AWS X-Ray Go SDK will make sure there will be a FacadeSegment inside lambda context so that customers are able to instrument their application successfully for subsegments generation case which including `Capture`, `HTTP Client`, `AWS`, `SQL` and `Custom Subsegments` usage.
```
yogiraj07: 1. "send them" -> "send them to X-Ray service" , probably this will be more clear.
2. nitpick : "including" should be changed to "includes"

PR #129, opened 2019-06-28
remove glide support

luluzhao:
*Description of changes:*
Remove glide support


luluzhao:
@yogiraj07 it won't impact customer since glide is just a dependency management(basically downloading the deps and track the version). At beginning go community didn't have a proper dependency management tool so that we choose glide at day one. However, since go community provides tools for us to use starting go version 1.9 and we already supported it, we should remove the glide. When we do a release, I will mention it in our changelog.

Issue #136, opened 2019-07-31
Data race in Rule/Target Pollers


yogiraj07:
# Issue
Data race condition
```
WARNING: DATA RACE
Read at 0x00c0000ae548 by goroutine 11:
  github.com/aws/aws-xray-sdk-go/strategy/sampling.(*CentralizedStrategy).snapshots()
      /Users/awati/go/src/github.com/aws/aws-xray-sdk-go/strategy/sampling/centralized.go:447 +0x8b
  github.com/aws/aws-xray-sdk-go/strategy/sampling.(*CentralizedStrategy).refreshTargets()
      /Users/awati/go/src/github.com/aws/aws-xray-sdk-go/strategy/sampling/centralized.go:367 +0x9e
  github.com/aws/aws-xray-sdk-go/strategy/sampling.(*CentralizedStrategy).startTargetPoller.func1()
      /Users/awati/go/src/github.com/aws/aws-xray-sdk-go/strategy/sampling/centralized.go:241 +0xf3

Previous write at 0x00c0000ae548 by goroutine 9:
  github.com/aws/aws-xray-sdk-go/strategy/sampling.(*CentralizedManifest).createUserRule()
      /Users/awati/go/src/github.com/aws/aws-xray-sdk-go/strategy/sampling/centralized_sampling_rule_manifest.go:134 +0xac3
  github.com/aws/aws-xray-sdk-go/strategy/sampling.(*CentralizedManifest).putRule()
      /Users/awati/go/src/github.com/aws/aws-xray-sdk-go/strategy/sampling/centralized_sampling_rule_manifest.go:75 +0x1cd
  github.com/aws/aws-xray-sdk-go/strategy/sampling.(*CentralizedStrategy).refreshManifest()
      /Users/awati/go/src/github.com/aws/aws-xray-sdk-go/strategy/sampling/centralized.go:321 +0x72e
  github.com/aws/aws-xray-sdk-go/strategy/sampling.(*CentralizedStrategy).startRulePoller.func1()
      /Users/awati/go/src/github.com/aws/aws-xray-sdk-go/strategy/sampling/centralized.go:209 +0x3c

Goroutine 11 (running) created at:
  github.com/aws/aws-xray-sdk-go/strategy/sampling.(*CentralizedStrategy).startTargetPoller()
      /Users/awati/go/src/github.com/aws/aws-xray-sdk-go/strategy/sampling/centralized.go:235 +0x4c
[...]

Issue #118, opened 2019-06-06
Data race when using aws clients


PaulLiang1:
Reproduced on:
- go version: 1.11.4
- goos: linux
- goarch: amd64
- aws-xray-sdk-go@v1.0.0-rc.11 
- aws-sdk-go@v1.17.12

when running the application with `-race` flag on it spits warning:
```
app_1   | ==================
app_1   | WARNING: DATA RACE
app_1   | Read at 0x00c00059fd10 by goroutine 91:
app_1   |   reflect.Value.Bool()
app_1   |       /usr/local/go/src/reflect/value.go:254 +0x51
app_1   |   encoding/json.isEmptyValue()
app_1   |       /usr/local/go/src/encoding/json/encode.go:319 +0x1d0
app_1   |   encoding/json.(*structEncoder).encode()
app_1   |       /usr/local/go/src/encoding/json/encode.go:636 +0x344
app_1   |   encoding/json.(*structEncoder).encode-fm()
app_1   |       /usr/local/go/src/encoding/json/encode.go:661 +0x7b
app_1   |   encoding/json.(*ptrEncoder).encode()
app_1   |       /usr/local/go/src/encoding/json/encode.go:788 +0x135
app_1   |   encoding/json.(*ptrEncoder).encode-fm()
app_1   |       /usr/local/go/src/encoding/json/encode.go:793 +0x7b
app_1   |   encoding/json.(*encodeState).reflectValue()
app_1   |       /usr/local/go/src/encoding/json/encode.go:333 +0x93
app_1   |   encoding/json.(*encodeState).marshal()
app_1   |       /usr/local/go/src/encoding/json/encode.go:305 +0xad
app_1   |   encoding/json.Marshal()
app_1   |       /usr/local/go/src/encoding/json/encode.go:160 +0x73
app_1   |   github.com/aws/aws-xray-sdk-go/xray.packSegments.func1()
app_1   |       /go/pkg/mod/github.com/aws/aws-xray-sdk-go@v1.0.0-rc.11/xray/default_emitter.go:101 +0x318
app_1   |   github.com/aws/aws-xray-sdk-go/xray.packSegments()
app_1   |       /go/pkg/mod/github.com/aws/aws-xray-sdk-go@v1.0.0-rc.11/xray/default_emitter.go:107 +0x29b
app_1   |   github.com/aws/aws-xray-sdk-go/xray.(*DefaultEmitter).Emit()
[...]

PR #140, opened 2019-08-06
Fixing bi-directional locking for parent-child segments and modifying lock on Segment struct to RWMutex

yogiraj07:
*Issue #118 , #136:*

*Description of changes:*
1. Bi-directional locking on parent/child (Sub)Segment leads to deadlocks
and data race conditions. This commit fixes this issue by making locking
uni-directional.

2. Changed lock on `Segment` struct to `RWMutex`. Added thread safe block for closing `connect` subsegment in aws.go. Changed lock to read access on `SafeInProgress()`.

### DIFF
--- a/xray/aws.go
+++ b/xray/aws.go
@@ -93,7 +93,7 @@ var xRayAfterSendHandler = request.NamedHandler{
 	Fn: func(r *request.Request) {
 		curseg := GetSegment(r.HTTPRequest.Context())
 
-		if curseg != nil && curseg.Name == "attempt" {
+		if curseg != nil && curseg.getName() == "attempt" {
 			// An error could have prevented the connect subsegment from closing,
 			// so clean it up here.
 			curseg.RLock()
--- a/xray/aws.go
+++ b/xray/aws.go
@@ -93,14 +93,15 @@ var xRayAfterSendHandler = request.NamedHandler{
 	Fn: func(r *request.Request) {
 		curseg := GetSegment(r.HTTPRequest.Context())
 
-		if curseg != nil && curseg.getName() == "attempt" {
+		if curseg != nil && curseg.Name == "attempt" {
 			// An error could have prevented the connect subsegment from closing,
 			// so clean it up here.
 			curseg.RLock()
-			s := curseg.rawSubsegments
+			temp := make([]*Segment, len(curseg.rawSubsegments))
+			copy(temp, curseg.rawSubsegments)
 			curseg.RUnlock()
 
-			for _, subsegment := range s {
+			for _, subsegment := range temp {
 				if subsegment.getName() == "connect" && subsegment.safeInProgress() {
 					subsegment.Close(nil)
 					return
--- a/xray/aws.go
+++ b/xray/aws.go
@@ -96,8 +96,12 @@ var xRayAfterSendHandler = request.NamedHandler{
 		if curseg != nil && curseg.Name == "attempt" {
 			// An error could have prevented the connect subsegment from closing,
 			// so clean it up here.
-			for _, subsegment := range curseg.rawSubsegments {
-				if subsegment.Name == "connect" && subsegment.safeInProgress() {
+			curseg.RLock()
+			s := curseg.rawSubsegments
+			curseg.RUnlock()
+
+			for _, subsegment := range s {
+				if subsegment.getName() == "connect" && subsegment.safeInProgress() {
 					subsegment.Close(nil)
 					return
 				}
--- a/xray/default_emitter.go
+++ b/xray/default_emitter.go
@@ -56,6 +56,7 @@ func (de *DefaultEmitter) refresh(raddr *net.UDPAddr) (err error) {
 }
 
 // Emit segment or subsegment if root segment is sampled.
+// seg has a write lock acquired by the caller.
 func (de *DefaultEmitter) Emit(seg *Segment) {
 	if seg == nil || !seg.ParentSegment.Sampled {
 		return
@@ -85,6 +86,7 @@ func (de *DefaultEmitter) Emit(seg *Segment) {
 	}
 }
 
+// seg has a write lock acquired by the caller.
 func packSegments(seg *Segment, outSegments [][]byte) [][]byte {
 	trimSubsegment := func(s *Segment) []byte {
 		ss := globalCfg.StreamingStrategy()
--- a/xray/default_emitter.go
+++ b/xray/default_emitter.go
@@ -98,17 +98,22 @@ func packSegments(seg *Segment, outSegments [][]byte) [][]byte {
 			cb := ss.StreamCompletedSubsegments(s)
 			outSegments = append(outSegments, cb...)
 		}
-		b, _ := json.Marshal(s)
+		b, err := json.Marshal(s)
+		if err != nil {
+			logger.Errorf("JSON error while marshalling (Sub)Segment: %v", err)
+		}
 		return b
 	}
 
 	for _, s := range seg.rawSubsegments {
+		s.Lock()
 		outSegments = packSegments(s, outSegments)
 		if b := trimSubsegment(s); b != nil {
 			seg.Subsegments = append(seg.Subsegments, b)
 		}
+		s.Unlock()
 	}
-	if seg.parent == nil {
+	if seg.isOrphan() {
 		if b := trimSubsegment(seg); b != nil {
 			outSegments = append(outSegments, b)
 		}
--- a/xray/default_streaming_strategy.go
+++ b/xray/default_streaming_strategy.go
@@ -11,17 +11,18 @@ package xray
 import (
 	"encoding/json"
 	"errors"
+	"sync/atomic"
 
 	"github.com/aws/aws-xray-sdk-go/internal/logger"
 )
 
-var defaultMaxSubsegmentCount = 20
+var defaultMaxSubsegmentCount uint32 = 20
 
 // DefaultStreamingStrategy provides a default value of 20
 // for the maximum number of subsegments that can be emitted
 // in a single UDP packet.
 type DefaultStreamingStrategy struct {
-	MaxSubsegmentCount int
+	MaxSubsegmentCount uint32
 }
 
 // NewDefaultStreamingStrategy initializes and returns a
@@ -37,14 +38,15 @@ func NewDefaultStreamingStrategyWithMaxSubsegmentCount(maxSubsegmentCount int) (
 	if maxSubsegmentCount <= 0 {
 		return nil, errors.New("maxSubsegmentCount must be a non-negative integer")
 	}
-	return &DefaultStreamingStrategy{MaxSubsegmentCount: maxSubsegmentCount}, nil
+	c := uint32(maxSubsegmentCount)
+	return &DefaultStreamingStrategy{MaxSubsegmentCount: c}, nil
 }
 
 // RequiresStreaming returns true when the number of subsegment
 // children for a given segment is larger than MaxSubsegmentCount.
 func (dSS *DefaultStreamingStrategy) RequiresStreaming(seg *Segment) bool {
 	if seg.ParentSegment.Sampled {
-		return seg.ParentSegment.totalSubSegments > dSS.MaxSubsegmentCount
+		return atomic.LoadUint32(&seg.ParentSegment.totalSubSegments) > dSS.MaxSubsegmentCount
 	}
 	return false
 }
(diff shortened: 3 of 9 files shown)


xray/default_emitter.go
```diff
@@ -98,17 +100,22 @@ func packSegments(seg *Segment, outSegments [][]byte) [][]byte {
 			cb := ss.StreamCompletedSubsegments(s)
 			outSegments = append(outSegments, cb...)
 		}
-		b, _ := json.Marshal(s)
+		b, err := json.Marshal(s)
+		if err != nil {
+			logger.Errorf("JSON error while marshalling (Sub)Segment: %v", err)
```
luluzhao: I think we never check the error when we marshal the segment since the segment is what we create. If we have this error, what should we do? I mean what is the use case for catching and logging this error from the customers' perspective?

xray/segment.go
```diff
@@ -384,7 +418,6 @@ func (sub *Segment) beforeEmitSubsegment(seg *Segment) {
 	sub.ParentID = seg.ID
 	sub.Type = "subsegment"
 	sub.RequestWasTraced = seg.RequestWasTraced
-	sub.parent = nil
```
luluzhao: Why we remove this line? Does that mean the subsegment after completing streaming strategy will still have a parent?

xray/segment.go
```diff
@@ -286,34 +296,56 @@ func (seg *Segment) RemoveSubsegment(remove *Segment) bool {
...
+func (seg *Segment) isOrphan() bool {
+	return seg.parent == nil || seg.Type == "subsegment"
+}
+
 func (seg *Segment) emit() {
 	seg.ParentSegment.GetConfiguration().Emitter.Emit(seg)
 }
 
-func (seg *Segment) handleContextDone() {
+func (seg *Segment)    handleContextDone() {
```
luluzhao: Could we move the extra indent before the method?

xray/segment.go
```diff
@@ -286,34 +296,56 @@ func (seg *Segment) RemoveSubsegment(remove *Segment) bool {
...
 	seg.Lock()
-	defer seg.Unlock()
-
 	seg.ContextDone = true
 	if !seg.InProgress && !seg.Emitted {
-		seg.flush()
+		s := seg
+		for {
+			if s.flush() {
```
luluzhao: Could you explain why we create a for loop here and the logic behind this?

xray/aws.go
```diff
@@ -96,8 +96,12 @@ var xRayAfterSendHandler = request.NamedHandler{
 		if curseg != nil && curseg.Name == "attempt" {
 			// An error could have prevented the connect subsegment from closing,
 			// so clean it up here.
-			for _, subsegment := range curseg.rawSubsegments {
-				if subsegment.Name == "connect" && subsegment.safeInProgress() {
+			curseg.RLock()
+			s := curseg.rawSubsegments
```
luluzhao: Why we create a copy here instead of directly using it?

xray/default_emitter.go
yogiraj07: It is FYI for the customer regarding the marshaling error. Currently due to any reason, if there is marshaling error, say due to unacceptable subsegment name provided by the customer, the entire segment may fail due to marshaling. The customer may wonder why the Segment is not appearing on the X-Ray console. This log will help in debugging the issue.

xray/aws.go
yogiraj07: With regards to optimization, in the sense, we hit this code for every AWS SDK API call being traced. Since we are iterating over `rawSegments`, its better to have copy, so another Go routine, if present, can update same `Segment` with minimum lock overhead.

xray/segment.go
yogiraj07: Flushing performance emit operation. For given Segment/Subsegment, we check if we are able to flush? -> If it has parent don't flush, if it has no parent then flush (there are more condition, you can check inside flush()). 

If it doesn't flush, we traverse to it parent and check whether we are able to flush parent.

So the for loop traverse from current node to root segment and at each level it checks, whether the node is able to flush. If a node is valid and ready to flush, we flush entire subtree. 


For another comment why we have  removed `sub.parent = nil` in `beforeEmitSubsegment()` is the reason for this "new" flush logic which is bottom up traversal of `Segment` tree. If we set parent to nil for a subsegment, we will emit each subsegment for non-streaming mode inside `flush()`. Which shouldn't be the case. The SDK sends entire Segment document in non-streaming mode.

This won't affect streaming mode since, since the streaming mode function `StreamCompletedSubsegments` breaks a subtree on the length of subtree and not on parent nil check.

xray/segment.go
yogiraj07: Replied in previous comment, since its more appropriate to explain there.

xray/segment.go
yogiraj07: done

xray/segment.go
haotianw465: Isn't line 327 - 339 exactly the same as what you added on segment.go `Close()` method? Is it possible to re-use the code? I'm not super familiar with the entire code base but I'm wondering why there are two places doing exactly the same thing.

xray/segment.go
```diff
@@ -322,25 +354,27 @@ func (seg *Segment) flush() {
 			logger.Debugf("emit lambda subsegment named: %v", seg.Name)
 			seg.emit()
 		} else {
-			seg.parent.safeFlush()
+			return false
```
haotianw465: Looks like you removed the recursiveness to force the caller of `flush()` to handle it. That's probably one of the reasons you have duplicate traverse logic in multiple places. This is not very easy to maintain. I think this particular method wit the new change is the right way. A segment should not flush other segments inside its own locked method. But IMO there should be some central unit to control the whole streaming/emission instead of have a complicated for-loop everywhere on calling `flush()`.

xray/segment.go
```diff
@@ -322,25 +354,27 @@ func (seg *Segment) flush() {
...
-	seg.Lock()
+	seg.RLock()
 	b := seg.InProgress
-	seg.Unlock()
+	seg.RUnlock()
 	return b
 }
 
+// getName returns name of the segment. This method is thread safe.
+func (seg *Segment) getName() string {
```
haotianw465: What is the purpose of this? I get that you can wrap every single getter call with `Rlock()/Runlock()` without changing the correctness of the program. We still need to use locks wisely.

xray/segment.go
yogiraj07: Sure, reusing the code is a good point (explained in later comment). The `flush()` logic is called from 2 places in the code. [Close()](https://github.com/aws/aws-xray-sdk-go/blob/master/xray/segment.go#L250) and [handleContextDone()](https://github.com/aws/aws-xray-sdk-go/blob/master/xray/segment.go#L310). For later function, the context may be cancelled by the framework for a given incoming request due to abrupt termination of the request.  For this scenario, we have a sweeping `Segment` logic inside [handleContextDone](https://github.com/aws/aws-xray-sdk-go/blob/master/xray/segment.go#L310). Hence, 2 places calling flush().

xray/aws.go
yogiraj07: Sure,
This code is executed for AWS API call, traced by the AWS SDK handler. For a given AWS SDK call we have possibility of generating `connect` subsegment inside `attempt` subsegment. The structure looks like :
```
SegA
|-Dynamo(SubsegA)
|-|-attempt (SubSegB)
|---|----connect(SubSegC)
|---|----request(SubSegD)
|---|----response(SubSegE)
```
The above code generates a copy for all children of `attempt` subsegment and closes `connect` subsegment. Note that [rawSegments](https://github.com/aws/aws-xray-sdk-go/blob/master/xray/segment_model.go#L69) is a array of pointers to `Segment` instances. Hence even after we make copy, the locks held by this copy are on the same instances of `Subsegment` and are valid inside the loop. 

 `SubSegB.rawSegments` should be thread safe since another thread may populate children of `attempt` subsegment. 
The unit test written for aws sdk call produces a data race, where we are iterating children of `attempt` subsegment, however, `response` subsegment is concurrently populated leading to data race at `subsegment.Name` and `SubSegB.rawSegments`. Hence the access to name of the subsegment is made thread safe by calling `getName()` method and making a copy respectively. 

*Note* :`s := curseg.rawSubsegments` is not a deep copy . Hence in the updated revision, I have made a deep copy to prevent any data races w.r.t `SubSegB.rawSegments` being populated by another go routine while iterating in this code block.

xray/segment.go
yogiraj07: Good point, I have refactored logic. A new method `send()` is created which is  thread safe and takes care of sending current (Sub)Segment or parent if its ready to send. Going ahead `send()` should be called.

xray/segment.go
yogiraj07: I explained the use case in the previous comment. Currently this is called only by `aws.go` closing `connect` subsegment logic.

xray/default_streaming_strategy.go
```diff
@@ -64,12 +66,15 @@ func (dSS *DefaultStreamingStrategy) StreamCompletedSubsegments(seg *Segment) []
 		seg.Subsegments[len(seg.Subsegments)-1] = nil
 		seg.Subsegments = seg.Subsegments[:len(seg.Subsegments)-1]
 
-		seg.ParentSegment.totalSubSegments--
+		atomic.AddUint32(&seg.ParentSegment.totalSubSegments, ^uint32(0))
```
luluzhao: What does the `^` mean in front of the `uinit32`?

xray/default_emitter.go
luluzhao: If there is an error, the program will panic right? Why not we panic at here?

xray/default_streaming_strategy.go
```diff
@@ -64,12 +66,15 @@ func (dSS *DefaultStreamingStrategy) StreamCompletedSubsegments(seg *Segment) []
...
 
 		// Add extra information into child subsegment
 		child.Lock()
 		child.beforeEmitSubsegment(seg)
-		cb, _ := json.Marshal(child)
+		cb, err:= json.Marshal(child)
+		if err!= nil{
```
luluzhao: same here, if cb is nil, what will happen? We cannot just simply throw an error message here since customer might start spending time debugging this which should be their problem.

xray/segment.go
```diff
@@ -234,8 +232,7 @@ func NewSegmentFromHeader(ctx context.Context, name string, h *header.Header) (c
 // Close a segment.
 func (seg *Segment) Close(err error) {
 	seg.Lock()
-	defer seg.Unlock()
-	if seg.parent != nil {
+	if seg.parent != nil || seg.Type == "Subsegment" {
```
luluzhao: Why we add a condition `seg.Type == "Subsegment"` here? How to trigger this one? I think the stream strategy happened after the segment getting closed. I feel this condition will never be triggered since we do assigned `Subsegment` type during streaming strategy.

xray/segment.go
```diff
@@ -269,51 +268,81 @@ func (subseg *Segment) CloseAndStream(err error) {
...
 	subseg.beforeEmitSubsegment(subseg.parent)
-	subseg.Unlock()
-
 	subseg.emit()
 }
 
 // RemoveSubsegment removes a subsegment child from a segment or subsegment.
 func (seg *Segment) RemoveSubsegment(remove *Segment) bool {
 	seg.Lock()
-	defer seg.Unlock()
```
luluzhao: Why we remove defer here? defer will be called before return.

xray/default_emitter.go
yogiraj07: Current behavior - If there is error, b will be nil, and we don't stream that (Sub)Segment. 
So logging the error helps if there is any error. If we panic instead of error, this is "behavioral" change.

xray/default_streaming_strategy.go
yogiraj07: This means to decrement the value by 1 : [reference](https://golang.org/src/sync/atomic/doc.go?s=3850:3905#L88)

xray/default_streaming_strategy.go
yogiraj07: Same as previous reply, if cb is nil - we don't stream the Subsegment. No error/panic happens. We shouldn't panic everywhere for possible marshaling error. IMHO, logging is the right strategy.

xray/segment.go
yogiraj07: For fine granular locking - decrementing `totalSubSegments` count is thread safe now and hence prior to that we unlock the subsegment : [reference](https://github.com/aws/aws-xray-sdk-go/pull/140/files#diff-4178d11b198e2b2e67f13586d5f83fc5R288)

xray/segment.go
yogiraj07: Good point. This check is unnecessary and I have removed it.

Issue #131, opened 2019-07-11
Password not removed when connecting to MySQL


zijima:
Hi,

I'm trying to open a new SQL connection this [required DSN format:](https://github.com/go-sql-driver/mysql#dsn-data-source-name): 

> user:password@(host:port)/database

However, the password section is not removed from connection string therefore i'm able to see the full DSN, including pwd, when i open traces in AWS XRay service Console. 

Looks like [current code](https://github.com/aws/aws-xray-sdk-go/blob/master/xray/sql_test.go#L115) can handle case like:

> user:password@host:port/database

But not the one with `()`

Source code also mentioned current [logic is based on surveys from 2017. ](https://github.com/aws/aws-xray-sdk-go/blob/master/xray/sql.go#L78)

Any chance you can include parsing the DSN format i needed?

thanks

luluzhao:
@zijima thank you for contacting us regarding your issue. We will add this into our backlog and in the meantime, A PR is always welcomed.

zijima:
It turns out that my DSN failed [url.parse](https://github.com/aws/aws-xray-sdk-go/blob/master/xray/sql.go#L42) because my password has characters like `[` and `]` which is treated as [invalid user info](https://github.com/golang/go/blob/master/src/net/url/url.go#L1102)  :(

i have a suggestion to handle the `net/url: invalid userinfo` error and remove password from the url

bhautikpip:
Hi @zijima,

We have recently merged one PR which solves this issue (https://github.com/aws/aws-xray-sdk-go/pull/169/files#diff-5bba2290697477468cb32ac08c47bd3aR733). So, I am closing this issue for now. Feel free to reopen an issue if you see any more issues.

PR #160, opened 2019-12-04
Custom SQL password sanitizing support

shayokneogi:
*Issue #, if available:*
#131 but more general
*Description of changes:*
Adds a method by which users may pass their own function for sanitizing the password from their database connection string when the `xray.SQL` method is not able to properly remove the password. This allows users to create their own methods to take care of DSN edge cases such as when the DSN contains a `[`, `(`, or the host protocol (i.e. `tcp`).


willarmiros:
Hi @shayokneogi,

Thank you for your contribution. We acknowledge the need for custom sanitizers to be applied to non-conformative DSNs as you described. However, we believe the best customer experience is to make any custom sanitization mutually exclusive with the existing default sanitization in order to not lead to any confusing outcomes from double sanitization. 

The approach we'd like to see is to implement a new method (e.g. `SQLwithSanitizer`) that is identical to the existing `SQL` method except that:
1. It takes the sanitizer method as a third parameter.
2. It does not perform any of the default sanitization done in the current `SQL` method.

I'd also appreciate some unit tests to verify this behavior. Please let me know if you have differing design opinions, and thank you again.

willarmiros:
Closing in favor of #191.

PR #161, opened 2019-12-05
context missing environment varibale fix

bhautikpip:
Tested this change on lambda by setting up the environment variable and also tested this locally. 

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/xray/config.go
+++ b/xray/config.go
@@ -80,9 +80,16 @@ func newGlobalConfig() *globalConfig {
 	}
 	ret.emitter = emt
 
-	cm := ctxmissing.NewDefaultRuntimeErrorStrategy()
-
-	ret.contextMissingStrategy = cm
+	cms := os.Getenv("AWS_XRAY_CONTEXT_MISSING")
+	if cms != "" {
+		if cms == ctxmissing.RuntimeErrorStrategy {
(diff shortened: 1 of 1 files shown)


xray/config.go
```diff
@@ -80,9 +80,16 @@ func newGlobalConfig() *globalConfig {
 	}
 	ret.emitter = emt
 
-	cm := ctxmissing.NewDefaultRuntimeErrorStrategy()
-
-	ret.contextMissingStrategy = cm
+	cms := os.Getenv("AWS_XRAY_CONTEXT_MISSING")
+	if cms != "" {
```
willarmiros: I believe this leaves the context missing strategy undefined if the environment variable isn't set? This is acceptable behavior in the `contextWithConfig` method because that *reads* a given, prefilled config object, however this method should be *setting* those fields to a well-defined value. Adding an `else` statement to set the CMS to runtime error should do the trick.

bhautikpip: That's a good point. I have added that logic earlier but what happens is lambda automatically sets environment variable (AWS_XRAY_CONTEXT_MISSING) by default (means when we don't set the environment variable) as LOG_ERROR so that else loop will never get executed.

bhautikpip: I think it would be good for non lambda customers also so yeah i will add that in next commit. Thanks for the suggestion.

Issue #98, opened 2019-04-02
Conn( ) support


pjebs:
Can xray please support `Conn`: https://golang.org/pkg/database/sql/#DB.Conn

https://godoc.org/github.com/aws/aws-xray-sdk-go/xray#DB

This is of little value without support for obtaining an exclusive connection.

luluzhao:
Hello @pjebs, thank you so much for contacting us. We will add this one into our backlog. In the meantime, we also accept PRs. If you would like to send us PR, we are happy to help you.

Issue #13, opened 2017-09-05
Restore *sql.DB compliance for Go 1.8+


itsjamie:
The method signature for xray.DB for Go 1.8 seems like it should implement the context methods the same as database/sql to make it easy to interop with other packages, currently, the context is passed to the Exec/Query/QueryRow functions and not their {*}Context versions.

itsjamie:
For example, #8 that is around support for sqlx, one solution here would be by simply changing the implementation for Go 1.8+ to be a drop-in replacement for database/sql, so Exec, Begin, etc would all be changed backed to their default method signatures, and if used would not provide any tracing information.

Personally, I'd prefer to keep drop-in support and allow to use all the existing tools that wrap database/sql, rather than having to make a choice between the two.

Issue #8, opened 2017-09-03
Support sqlx package


dlsniper:
Hi,

Thank you for creating this.

I would like to request adding support for ` github.com/jmoiron/sqlx ` in addition to ` database/sql `.
I'd be happy to provide the implementation if you'd be ok with it.
Looking forward for your reply.

itsjamie:
Hello @dlsniper, rather than adding support for the direct `sqlx.DB` type, what about changing this so it that it will accept an interface that `sql.DB` sastifies?

I've referenced #13 here, but I was thinking that if context is necessary, since Go is only supporting Go 1.8, Go 1.9 with the release of 1.9, for a new library it seems like it could only support back to Go 1.8.

Of course, AWS employees would have to comment, but then we could restore the `*sql.DB` compliant structure of the `xray.DB` type, and provide tracing information from there.

Then I don't think you would have to provide a separate implementation for the sqlx package.

dlsniper:
As far as I'm aware there is no interface for ` database/sql.*DB `.
And that would not be possible to begin with since there are methods on ` sqlx.*DB ` that are not present on ` sql.*DB `.

itsjamie:
Correct, there isn't an interface already defined for the `database/sql.*DB` package, however, this package could be changed to accept an interface that "database/sql.*DB" and sqlx.DB both implemented.

Since sqlx.DB is a superset of functionality on top of the existing package, similar to how this library is built now, I believe that a sqlx.DB could be passed if an interface was defined here.

itsjamie:
For example, https://play.golang.org/p/Giy1lbE1W3

itsjamie:
The real issue with that approach would be that this library currently isn't a drop-in replacement for database/sql like sqlx because of the < Go 1.8 support, the type signatures for most of the calls are different to require contexts.

dlsniper:
> because of the < Go 1.8 support

The official support for Go < 1.8 has ended as soon as Go 1.9 arrived. As such, when this library hits 1.0 I don't see any reason on why this shouldn't support Go 1.8+.

The problem with the interface approach is that if anything changes and a function is added, this package is then forced to break compatibility.

I'm working to show a different way to solve this, will take me a bit longer.

itsjamie:
One approach I was considering was creating a proposal for a `sqltrace` package similar to `httptrace` for the stdlib.

https://groups.google.com/forum/#!topic/golang-sql/nGh4Qn5KxCA

[...]

PR #169, opened 2020-01-11
proposal: implement custom driver for xray instead of xray.DB

shogo82148:
*Issue #, if available:*

Fixes https://github.com/aws/aws-xray-sdk-go/issues/98, https://github.com/aws/aws-xray-sdk-go/issues/13 and https://github.com/aws/aws-xray-sdk-go/issues/8

*Description of changes:*

Currently, the interface of `*xray.DB` is not compatible with `*database/sql.DB`.
So, it is hard to use some database utilities such as https://github.com/jmoiron/sqlx (ref. https://github.com/aws/aws-xray-sdk-go/issues/8), and we have to rewrite large amount our code.

I implemented the xray wrapper as a [custom SQL driver](https://golang.org/pkg/database/sql/driver/).
It can be used as a drop-in replacement because the `database/sql` package hides implementation differences. (fixes https://github.com/aws/aws-xray-sdk-go/issues/13)
Of course, `DB.Conn` https://golang.org/pkg/database/sql/#DB.Conn  is supported. (ref. https://github.com/aws/aws-xray-sdk-go/issues/98)

And more, we can trace more detail behavior (e.g. creating a new connection, resetting a session).


bhautikpip:
Hi @shogo82148,

Thanks for the PR. I will review it. Since this is a big one might take some time to get done with testing.

shogo82148:
My memo:

this work is based on https://github.com/shogo82148/go-sql-proxy
and I need to port https://github.com/shogo82148/go-sql-proxy/pull/42 before merging.

bhautikpip:
Hi @shogo82148,

Would you mind resolving the branch conflict ? Moreover, are you still going to work on this PR or should I review it ?

shogo82148:
> Would you mind resolving the branch conflict ?

Sure.

> are you still going to work on this PR or should I review it ?

I'm still going work, but any feedback about its design is welcome.

bhautikpip:
> > Would you mind resolving the branch conflict ?
> 
> Sure.
> 
> > are you still going to work on this PR or should I review it ?
> 
> I'm still going work, but any feedback about its design is welcome.

I think looks like a good design. I think SQLcontext is the right design approach if we want to support both.

bhautikpip:
I think there is still some data race happening in this implementation too. I looked through the debug results.

shogo82148:
@bhautikpip Can you review this pull request, please.

`reviewdog / golangci-lint` fails because of deprecated SQL functions.
They are need to support legacy SQL drivers, so please ignore these warnings.

xray/sql.go
```diff
@@ -9,290 +9,124 @@
...
-		db.driverVersion = t.PkgPath()
-	}
-
-	return db, nil
+	return &DB{DB: db}, nil
 }
 
 // DB copies the interface of sql.DB but adds X-Ray tracing.
 // It must be created with xray.SQL.
 type DB struct {
```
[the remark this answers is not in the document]
sthulb: I'm not sure we should care about backwards compat given officially we're not at v1.x

xray/sql_context.go
```diff
@@ -0,0 +1,725 @@
...
+	if connCtx, ok := conn.Conn.(driver.ConnBeginTx); ok {
+		tx, err = connCtx.BeginTx(ctx, opts)
+	} else {
+		if opts.Isolation != driver.IsolationLevel(sql.LevelDefault) {
+			return nil, errors.New("xray: driver does not support non-default isolation level")
+		}
+		if opts.ReadOnly {
+			return nil, errors.New("xray: driver does not support read-only transactions")
+		}
+		tx, err = conn.Conn.Begin()
```
sthulb: We should probably handle these deprecation notices

xray/sql_context.go
```diff
@@ -0,0 +1,725 @@
...
+		})
+	}
+	return rows, err
+}
+
+func (conn *driverConn) Close() error {
+	return conn.Conn.Close()
+}
+
+// copied from sql/driver/convert.go
```
sthulb: sql/driver/convert.go – is from this repo or else where? If elsewhere, do we have to care about their license?

xray/sql.go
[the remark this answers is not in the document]
shogo82148: Yeah, that's right.
@bhautikpip Can I remove it now?

xray/sql_context.go
shogo82148: it is from https://github.com/golang/go/blob/e6ebbe0d20fe877b111cf4ccf8349cba129d6d3a/src/database/sql/convert.go#L93-L99

xray/sql_context.go
shogo82148: No, we shouldn't. these notices are false positives.
We check the driver implements the new `driver.ConnBeginTx` interface in L152.

```go
if connCtx, ok := conn.Conn.(driver.ConnBeginTx); ok {
```

And it is fallback for legacy drivers that don't implement `driver.ConnBeginTx` interface.
It is same behavior as Go's standard library https://github.com/golang/go/blob/e6ebbe0d20fe877b111cf4ccf8349cba129d6d3a/src/database/sql/ctxutil.go#L125

xray/sql.go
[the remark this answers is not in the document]
bhautikpip: Yes! I think we should remove this since it's not at v1.x

xray/sql_go110.go
```diff
@@ -0,0 +1,155 @@
...
+
+import (
+	"context"
+	"database/sql/driver"
+	"sync"
+)
+
+// SQLConnector wraps the connector, and traces SQL executions.
+// Unlike SQLContext, SQLConnector doesn't filter the password of the dsn.
+// So, you have to filter the password before passing the dns to SQLConnector.
```
bhautikpip: typo: dsn to SQLConnector

shogo82148: fixed!

xray/sql_context.go
```diff
@@ -0,0 +1,725 @@
...
+		switch c {
+		case ':', delimiter:
+			flush()
+		case '=':
+			tokStr := strings.ToLower(tok.String())
+			isPassword = `password=` == tokStr || `pwd=` == tokStr
+			if b, err := buf.ReadByte(); err == nil && b == '{' {
+				inBraces = true
+			}
+			buf.UnreadByte()
```
bhautikpip: I think we should check for error return value as suggested by reviewdog

xray/sql_context.go
```diff
@@ -0,0 +1,725 @@
...
+		case '}':
+			b, err := buf.ReadByte()
+			if err != nil {
+				break
+			}
+			if b == '}' {
+				tok.WriteByte(b)
+			} else {
+				inBraces = false
+				buf.UnreadByte()
```
bhautikpip: I think we should check for error return value as suggested by reviewdog

xray/sql_go110.go
```diff
@@ -0,0 +1,155 @@
...
+	attr *dbAttribute
+}
+
+func (c *driverConnector) Connect(ctx context.Context) (driver.Conn, error) {
+	var rawConn driver.Conn
+	attr, err := c.getAttr(ctx)
+	if err != nil {
+		return nil, err
+	}
+	err = Capture(ctx, attr.dbname, func(ctx context.Context) error {
```
bhautikpip: We should name the subsegment properly otherwise all the subsegments will have the same name because we pass attr.dbname as the name of the subsegment.

xray/sql_go110.go
```diff
@@ -0,0 +1,155 @@
...
+		driver:    d,
+		name:      dsn,
+		filtered:  true,
+		// initialized attr lazy because we have no context here.
+	}
+}
+
+func (conn *driverConn) ResetSession(ctx context.Context) error {
+	if sr, ok := conn.Conn.(driver.SessionResetter); ok {
+		return Capture(ctx, conn.attr.dbname, func(ctx context.Context) error {
```
bhautikpip: name the subsegment.

xray/sql_context.go
```diff
@@ -0,0 +1,725 @@
...
+	return conn, nil
+}
+
+type driverConn struct {
+	driver.Conn
+	attr *dbAttribute
+}
+
+func (conn *driverConn) Ping(ctx context.Context) error {
+	return Capture(ctx, conn.attr.dbname, func(ctx context.Context) error {
```
bhautikpip: name the subsegment "ping"

shogo82148:
I'll fix it.

the `database/sql` package resets sessions in background goroutine.
https://github.com/golang/go/blob/843fec1c7d75cac3f76620e79f1680d8f058c501/src/database/sql/sql.go#L1073
so we can't capture ResetSession.
it doesn't reproduce in tests, because `DATA-DOG/go-sqlmock` doesn't implement `driver.SessionResetter`.

xray/sql_context.go
```diff
@@ -0,0 +1,725 @@
...
+func (conn *driverConn) ExecContext(ctx context.Context, query string, args []driver.NamedValue) (driver.Result, error) {
+	execer, ok := conn.Conn.(driver.Execer)
+	if !ok {
+		return nil, driver.ErrSkip
+	}
+
+	var err error
+	var result driver.Result
+	if execerCtx, ok := conn.Conn.(driver.ExecerContext); ok {
+		err = Capture(ctx, conn.attr.dbname, func(ctx context.Context) error {
```
bhautikpip: Name the subsegment.

xray/sql_context.go
```diff
@@ -0,0 +1,725 @@
...
+		select {
+		default:
+		case <-ctx.Done():
+			return nil, ctx.Err()
+		}
+		dargs, err0 := namedValuesToValues(args)
+		if err0 != nil {
+			return nil, err0
+		}
+		err = Capture(ctx, conn.attr.dbname, func(ctx context.Context) error {
```
bhautikpip: Name the subsegment.

xray/sql_context.go
```diff
@@ -0,0 +1,725 @@
...
+func (conn *driverConn) QueryContext(ctx context.Context, query string, args []driver.NamedValue) (driver.Rows, error) {
+	queryer, ok := conn.Conn.(driver.Queryer)
+	if !ok {
+		return nil, driver.ErrSkip
+	}
+
+	var err error
+	var rows driver.Rows
+	if queryerCtx, ok := conn.Conn.(driver.QueryerContext); ok {
+		err = Capture(ctx, conn.attr.dbname, func(ctx context.Context) error {
```
bhautikpip: Name the subsegment.

xray/sql_context.go
```diff
@@ -0,0 +1,725 @@
...
+		select {
+		default:
+		case <-ctx.Done():
+			return nil, ctx.Err()
+		}
+		dargs, err0 := namedValuesToValues(args)
+		if err0 != nil {
+			return nil, err0
+		}
+		err = Capture(ctx, conn.attr.dbname, func(ctx context.Context) error {
```
bhautikpip: Name the subsegment.

xray/sql_context.go
```diff
@@ -0,0 +1,725 @@
...
+
+func (stmt *driverStmt) Exec(args []driver.Value) (driver.Result, error) {
+	panic("not supported")
+}
+
+func (stmt *driverStmt) ExecContext(ctx context.Context, args []driver.NamedValue) (driver.Result, error) {
+	var result driver.Result
+	var err error
+	if execerContext, ok := stmt.Stmt.(driver.StmtExecContext); ok {
+		err = Capture(ctx, stmt.attr.dbname, func(ctx context.Context) error {
```
bhautikpip: Name the subsegment.

xray/sql_context.go
```diff
@@ -0,0 +1,725 @@
...
+		select {
+		default:
+		case <-ctx.Done():
+			return nil, ctx.Err()
+		}
+		dargs, err0 := namedValuesToValues(args)
+		if err0 != nil {
+			return nil, err0
+		}
+		err = Capture(ctx, stmt.attr.dbname, func(ctx context.Context) error {
```
bhautikpip: Name the subsegment.

xray/sql_context.go
```diff
@@ -0,0 +1,725 @@
...
+
+func (stmt *driverStmt) Query(args []driver.Value) (driver.Rows, error) {
+	panic("not supported")
+}
+
+func (stmt *driverStmt) QueryContext(ctx context.Context, args []driver.NamedValue) (driver.Rows, error) {
+	var result driver.Rows
+	var err error
+	if queryCtx, ok := stmt.Stmt.(driver.StmtQueryContext); ok {
+		err = Capture(ctx, stmt.attr.dbname, func(ctx context.Context) error {
```
bhautikpip: Name the subsegment.

xray/sql_context.go
```diff
@@ -0,0 +1,725 @@
...
+		select {
+		default:
+		case <-ctx.Done():
+			return nil, ctx.Err()
+		}
+		dargs, err0 := namedValuesToValues(args)
+		if err0 != nil {
+			return nil, err0
+		}
+		err = Capture(ctx, stmt.attr.dbname, func(ctx context.Context) error {
```
bhautikpip: Name the subsegment.

bhautikpip:
> I'll fix it.
> 
> the `database/sql` package resets sessions in background goroutine.
> https://github.com/golang/go/blob/843fec1c7d75cac3f76620e79f1680d8f058c501/src/database/sql/sql.go#L1073
> so we can't capture ResetSession.
> it doesn't reproduce in tests, because `DATA-DOG/go-sqlmock` doesn't implement `driver.SessionResetter`.

Yes. Makse sense!

xray/sql_go110.go
shogo82148: ok.
should I rename all subsegments(e.g. Ping, Exec, Query, etc.)?
currently, all subsegments have same name `attr.dbname`.

https://github.com/aws/aws-xray-sdk-go/blob/16d917bfd8d13242ae4f03807a10773a5e9cd468/xray/sql_go18.go#L33
https://github.com/aws/aws-xray-sdk-go/blob/16d917bfd8d13242ae4f03807a10773a5e9cd468/xray/sql_go18.go#L44

bhautikpip: Yes. That would be great!

bhautikpip:
> Hi @shogo82148,
> 
> I was testing sample code below with your branch.
> 
> ```
> import (
> 	"context"
> 	"fmt"
> 	"github.com/aws/aws-xray-sdk-go/xray"
> 	_ "github.com/go-sql-driver/mysql"
> )
> 
> func Handler()  {
> 	fmt.Println("Go MySQL Test")
> 	db, err := xray.SQLContext("mysql", "root:password@tcp(127.0.0.1:3306)/testdb")
> 	defer db.Close()
> 	if err != nil {
> 		return
> 	}
> 
> 	ctx, root := xray.BeginSegment(context.Background(), "Test SQL")
> 	if err := db.PingContext(ctx); err != nil {
> 		return
> 	}
> 	root.Close(nil)
> 
> 	fmt.Println("Successfully sent segments")
> }
> 
> func main() {
> 	Handler()
> }
> ```
> 
> I received segment can not be found. Basically when code hits (https://github.com/aws/aws-xray-sdk-go/pull/169/files#diff-66905444e94c2e73f12ac484eb8eb2d5R36) this line it runs Capture function and because of the empty context GetSegment() returns the empty parent and throws this panic. Let me know If I missed any configuration on above code.

Also, one more thing I have used above example and still seeing password in the subsegments. we should be able to strip password from dsn string right ?

shogo82148:
> name the subsegment "ping"

Should I change names of subsegments?
I got strange service map on the AWS managed console.

before changing: there is one Database::SQL "test" on my service map.
スクリーンショット 2020-02-10 17 56 42

after changing: there are three Database::SQL "connect", "exec" and "query" on my service map.
スクリーンショット 2020-02-10 18 01 23

the code is here https://gist.github.com/shogo82148/54fddd57a4d9ecb820a65547277643b8

-----

> Also, one more thing I have used above example and still seeing password in the subsegments. we should be able to strip password from dsn string right ?

yes, we should strip password from the dsn,
but now, it is not removed when connecting to MySQL. ref. https://github.com/aws/aws-xray-sdk-go/issues/131
I have two solutions.

-----

One is support for custom SQL password sanitizing ref. https://github.com/aws/aws-xray-sdk-go/pull/160 
Unfortunately, we can't use the implementation of #160, because the `*sql.DB` hides the implementation of the `*xray.DB`.
we have to rethink its implementation.

-----

The other hand is adding the sanitized DSN into `SQLContext` like the following.

```go
func SQLContext(driver, rawDSN, sanitizedDSN string) (*sql.DB, error) {
```

Its function signature is different from `sql.Open`'s one, but more safer than sanitizing.

bhautikpip:
Hi @shogo82148,

Yeah you are right. I think we should not change the name of the subsegments. I checked the previous implementation and subsegment name is derived from name of the database. Moreover, I think this PR should solve the SQL password sanitizing issue from dsn string. (https://github.com/aws/aws-xray-sdk-go/pull/191).  What do you think ?

shogo82148:
#191 looks good, but #191 removes the username if the password is not in the dns.
e.g.

https://github.com/aws/aws-xray-sdk-go/pull/169/commits/fb48a58f12472177cab1715e9ef3ae8027adf22a#diff-b87f5b0cbed57121f93560c7d82b570c

of course, it very rare case, but I'm not sure that it works well for other drivers.
So, I think custom SQL password sanitizing or sanitized DSN is sill needed. It makes the X-Ray SDK safer.

bhautikpip:
> #191 looks good, but #191 removes the username if the password is not in the dns.
> e.g.
> 
> [fb48a58#diff-b87f5b0cbed57121f93560c7d82b570c](https://github.com/aws/aws-xray-sdk-go/commit/fb48a58f12472177cab1715e9ef3ae8027adf22a#diff-b87f5b0cbed57121f93560c7d82b570c)
> 
> of course, it very rare case, but I'm not sure that it works well for other drivers.
> So, I think custom SQL password sanitizing or sanitized DSN is sill needed. It makes the X-Ray SDK safer.

Makes sense. I understood about your first approach but did not quite understand about this approach. 

The other hand is adding the sanitized DSN into SQLContext like the following.

`func SQLContext(driver, rawDSN, sanitizedDSN string) (*sql.DB, error) {`
Its function signature is different from sql.Open's one, but more safer than sanitizing.

Does this mean we also expect sanitized DSN from customer and use it to populate on X-Ray console and use rawDSN under the hood to make actual database connection ?

shogo82148:
> Does this mean we also expect sanitized DSN from customer and use it to populate on X-Ray console and use rawDSN under the hood to make actual database connection ?

Yes, that's right.

In additional, `xray.SQLConnector` expects sanitized DSN, because the connector has no information of actual database connection.

```go
func SQLConnector(dsn string, connector driver.Connector) driver.Connector {
```

bhautikpip:
Hi @shogo82148,

Got it! I have opened a PR for global sanitizing strategy design. I have also added SQL dsn sanitizer. Added sanitizing strategy in the global config. would you mind taking a look and provide your feedback ? (https://github.com/aws/aws-xray-sdk-go/pull/193).

shogo82148:
Sure.

But, I think SQL tracking and sanitizing are different problem.
should we discuss about sanitizing on another issue?

bhautikpip:
Yeah you're right.

xray/sql_context.go
```diff
@@ -0,0 +1,752 @@
...
+		}
+		if opts.ReadOnly {
+			return nil, errors.New("xray: driver does not support read-only transactions")
+		}
+		tx, err = conn.Conn.Begin()
+		if err == nil {
+			select {
+			default:
+			case <-ctx.Done():
+				tx.Rollback()
```
bhautikpip: We should check for error return value.

xray/sql_context.go
```diff
@@ -0,0 +1,752 @@
...
+			return nil, ctx.Err()
+		}
+		dargs, err0 := namedValuesToValues(args)
+		if err0 != nil {
+			return nil, err0
+		}
+		err = Capture(ctx, stmt.attr.dbname, func(ctx context.Context) error {
+			stmt.populate(ctx)
+			var err error
+			result, err = stmt.Stmt.Exec(dargs)
```
bhautikpip: can we disable deprecated warning reported by golangci-lint ?

xray/sql_context.go
shogo82148: we can't do anything if we get an error.
so, just ignore it here.

xray/sql_context.go
shogo82148: ignore warnings and merge this pull request please.
[reviewdog](https://github.com/reviewdog/reviewdog) will not report them next time.

bhautikpip: Hi @shogo82148,

I am going to merge this PR today. Just wanted to provide final suggestions. In one of the subsegment Sanitized Query field looks like this: 

Sanitized query | SELECT bar FROM foo WHERE bar = ? -- skipped

Do we want to add more information so that customer can understand why we get subsegment of skipped query and the same query gets traced successfully in next subsegment  ?

shogo82148: I have just committed https://github.com/aws/aws-xray-sdk-go/pull/169/commits/328f078137c50bb6735b34b16ed7666914af8873
how about it?

it comes from the message of `driver.ErrSkip`

```go
var ErrSkip = errors.New("driver: skip fast-path; continue as if unimplemented")
```

bhautikpip: Sorry for last min suggestions. I was wondering what is the point of tracing this error skip message as subsegments ? I think tracing this message will result in 2 subsegments of the same query which is redundant. Moreover, this does not add any instrumentation value as this msg just informs sql package should continue as if the optional interface is not implemented. What do you think ?

shogo82148: Yes, you are right. This subsegment may be redundant.
But there is no interface to remove existing subsegments, and we can't know whether we should add a new subsegment before calling `Exec`.

And the users can use this message as optimize hint.
e.g. the [interpolateParams](https://github.com/go-sql-driver/mysql#interpolateparams) option of go-sql-driver/mysql enables this fast-path. The users can know whether this option really works well via the trace.

sthulb:
Let me know when this is about to be merged and I'll give a final review.

xray/sql_context.go
bhautikpip: Okay Makes sense! Also, there was a customer ask to keep sql.go 
(previous implementation) and add a note that older (sql.go) implementation will be deprecated when we do Go-SDK GA. This will give more time to customers who use the existing implementation. Would you mind adding deleted file so when I merge this PR it will have the previous implementation same as the base branch.

bhautikpip: Okay. that's fair!

bhautikpip:
> Let me know when this is about to be merged and I'll give a final review.

Sure!

bhautikpip:
Hi @shogo82148,

I received one suggestion after discussing with customer who uses current SQL implementation. I think we should also keep current implementation of SQL because existing customer will have time to adept this change. The customer proposed we can add a note saying that older SQL implementation (without custom SQL driver) will be deprecated when we do GA. I think that is a fair call out.

shogo82148:
I reverted 96e9797 (that removes `xray.SQL`).

The implementation of `xray.SQL` is now a just wrapper for `xray.SQLContext`.
So, I think there is no need to import old `sql_test.go`.
And more `sql_test.go` has race condition and its tests sometimes fail.

xray/sql_context.go
```diff
@@ -0,0 +1,756 @@
...
+			}
+			if b == '}' {
+				tok.WriteByte(b)
+			} else {
+				inBraces = false
+				if err := buf.UnreadByte(); err != nil {
+					panic(err)
+				}
+			}
+		case '@':
```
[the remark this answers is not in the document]
bhautikpip: Hi @shogo82148,

would you mind adding this updated logic in this PR? It will cover the case when it does not have the password and won't strip the user.


```
case '@':
			if strings.Contains(res.String(), ":") {
				resLen := res.Len()
				if resLen > 0 && res.Bytes()[resLen-1] == ':' {
					res.Truncate(resLen - 1)
				}
				isPassword = true
				flush()
				res.WriteByte(c)
			}
```

xray/sqlcontext_test.go
```diff
@@ -0,0 +1,354 @@
...
+			want: "",
+		},
+
+		// test cases for https://github.com/go-sql-driver/mysql
+		{
+			in:   "user:password@tcp(localhost:5555)/dbname?tls=skip-verify&autocommit=true",
+			want: "user@tcp(localhost:5555)/dbname?tls=skip-verify&autocommit=true",
+		},
+
+		// got: "@tcp(localhost:5555)/dbname?tls=skip-verify&autocommit=true"
```
bhautikpip: you can uncomment this test cases after adding the above suggested logic.

bhautikpip: Also, I have merged this PR (https://github.com/aws/aws-xray-sdk-go/pull/191) to immediately unblock the customer. After we add this we can overwrite those changes.

bhautikpip:
You might need to remove this files https://github.com/aws/aws-xray-sdk-go/blob/master/xray/sql_prego111_test.go, https://github.com/aws/aws-xray-sdk-go/blob/master/xray/sql_go111_test.go.

bhautikpip:
Hi @sthulb ,

I am going to merge this PR today. would you mind taking a look ?

PR #170, opened 2020-01-14
Fix golint warnings

shogo82148:
*Issue #, if available:*

*Description of changes:*

There are some golangci-lint warnings.
https://travis-ci.org/aws/aws-xray-sdk-go/jobs/635402231

This pull request will fix them.


bhautikpip:
Hi @shogo82148,

Thanks for opening this PR. test cases are failing in lambda_go.test

shogo82148:
I think it is another problem, because the recent commit https://github.com/aws/aws-xray-sdk-go/pull/170/commits/62e6c364dad953231e133b1e4b5b690e74d8199b didn't change `TestLambdaSegmentEmit` and `lambda_go.test`.

-----

I'm worrying that https://github.com/aws/aws-xray-sdk-go/pull/170/commits/f6352d365b7335abeaf5353940ec79b2f137232d may not be acceptable.
because it changes the public field name of `sampling.Request` and breaks compatibility.
The direct use of `sampling.Request` is very rare case, so I think it doesn't have a large impact.
This change comes from https://github.com/golang/go/wiki/CodeReviewComments#initialisms .
Do you mind it?

shogo82148:
oh, the context key `"x-amzn-trace-id"` in AWS Lambda is hard coded in the aws-lambda-go.
https://github.com/aws/aws-lambda-go/blob/b5b7267d297de263cc5b61f8c37543daa9c95ffd/lambda/function.go#L65
I will revert it.

strategy/sampling/sampling_rule.go
```diff
@@ -201,6 +201,7 @@ type Rule struct {
 	*Properties
 }
 
+// Sample is ...
```
bhautikpip: Sample is used to provide sampling decision.

strategy/sampling/sampling_rule_manifest.go
```diff
@@ -85,20 +85,20 @@ func initSamplingRules(srm *RuleManifest) {
...
 func processManifest(srm *RuleManifest) error {
-	if nil == srm {
-		return errors.New("Sampling rule manifest must not be nil.")
+	if srm == nil {
+		return errors.New("sampling rule manifest must not be nil")
 	}
-	if 1 != srm.Version && 2 != srm.Version {
-		return errors.New(fmt.Sprintf("Sampling rule manifest version %d not supported.", srm.Version))
+	if srm.Version != 1 && srm.Version != 2 {
+		return fmt.Errorf("sampling rule manifest version %d not supported", srm.Version)
```
bhautikpip: What is the reason of using fmt.Errorf instead of errors.New in this case only ?

strategy/sampling/sampling_test.go
```diff
@@ -22,33 +22,10 @@ func TestNewLocalizedStrategy(t *testing.T) {
...
-        "url_path": "/checkout",
-        "fixed_target": 10,
-        "rate": 0.05
-       }
-	  ]
-	}`
-	goPath := os.Getenv("PWD")
-	testFile := goPath + "/test_rule.json"
-	f, err := os.Create(testFile)
+	testFile, err := filepath.Abs(filepath.Join("testdata", "rule-v1-sampling.json"))
```
bhautikpip: good callout!

xray/lambda_test.go
```diff
@@ -8,6 +8,9 @@ import (
 )
 
 func TestLambdaSegmentEmit(t *testing.T) {
+	// go-lint warns "should not use basic type string as key in context.WithValue",
+	// but it must be string type because the trace header comes from aws/aws-lambda-go.
```
bhautikpip: might want to remove this comment.

strategy/sampling/sampling_rule_manifest.go
shogo82148: because `errors.New("sampling rule manifest must not be nil")` has no format 'verbs' (e.g. `%d`)

bhautikpip: Makes sense!

bhautikpip:
Hi @shogo82148,

test case is failing because of this line (https://github.com/aws/aws-xray-sdk-go/pull/170/files#diff-4178d11b198e2b2e67f13586d5f83fc5R98). We might need to change this as URL and then we should be good to go with this changes.

bhautikpip:
Hi @shogo82148,

would you mind resolving the conflicts and add URL (segment.go - Line 98) instead of Url. This is PR is good to go other than that.

PR #177, opened 2020-01-16
create new TestDaemons for each test cases

shogo82148:
*Issue #, if available:*

I'm guessing this pull request fixes some broken tests. e.g. https://travis-ci.org/aws/aws-xray-sdk-go/jobs/637161870

*Description of changes:*

By default, `go test` run tests in parallel.
So multiple TestDaemons try to listen UDP on `127.0.0.1:2000` in almost same time,
and conflict each other.

> https://golang.org/cmd/go/#hdr-Compile_packages_and_dependencies
> ### Compile packages and dependencies
>
>     -p n
> 	    the number of programs, such as build commands or
> 	    test binaries, that can be run in parallel.
> 	    The default is the number of CPUs available.
> 

> https://golang.org/cmd/go/#hdr-Testing_flags
> ### Testing flags
>
>     -parallel n
>         Allow parallel execution of test functions that call t.Parallel.
>         The value of this flag is the maximum number of tests to run
>         simultaneously; by default, it is set to the value of GOMAXPROCS.
>         Note that -parallel only applies within a single test binary.
>         The 'go test' command may run tests for different packages
>         in parallel as well, according to the setting of the -p flag
>         (see 'go help build').
> 

I created new TestDaemons for each test cases to avoid this.

### DIFF
--- a/xray/aws_test.go
+++ b/xray/aws_test.go
@@ -1,3 +1,11 @@
+// Copyright 2017-2017 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License"). You may not use this file except in compliance with the License. A copy of the License is located at
+//
+//     http://aws.amazon.com/apache2.0/
+//
+// or in the "license" file accompanying this file. This file is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the specific language governing permissions and limitations under the License.
+
 package xray
 
 import (
@@ -20,7 +28,7 @@ func TestAWS(t *testing.T) {
 	// Runs a suite of tests against two different methods of registering
 	// handlers on an AWS client.
 
-	type test func(*testing.T, *lambda.Lambda)
+	type test func(context.Context, *TestDaemon, *testing.T, *lambda.Lambda)
 	tests := []struct {
 		name     string
 		test     test
@@ -67,23 +75,32 @@ func TestAWS(t *testing.T) {
 
 	// Run all combinations of constructors + tests.
 	for _, cons := range constructors {
+		cons := cons
 		t.Run(cons.name, func(t *testing.T) {
 			for _, test := range tests {
+				test := test
+				ctx, td := NewTestDaemon()
+				defer td.Close()
+
 				t.Run(test.name, func(t *testing.T) {
-					test.test(t, cons.constructor(fakeSession(t, test.failConn)))
+					session, cleanup := fakeSession(t, test.failConn)
+					defer cleanup()
+					test.test(ctx, td, t, cons.constructor(session))
 				})
 			}
 		})
 	}
 }
 
-func fakeSession(t *testing.T, failConn bool) *session.Session {
+func fakeSession(t *testing.T, failConn bool) (*session.Session, func()) {
 	cfg := &aws.Config{
 		Region:      aws.String("fake-moon-1"),
 		Credentials: credentials.NewStaticCredentials("akid", "secret", "noop"),
 	}
+
+	var ts *httptest.Server
 	if !failConn {
-		ts := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		ts = httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 			b := []byte(`{}`)
 			w.WriteHeader(http.StatusOK)
 			w.Write(b)
@@ -92,22 +109,28 @@ func fakeSession(t *testing.T, failConn bool) *session.Session {
 	}
 	s, err := session.NewSession(cfg)
 	assert.NoError(t, err)
-	return s
+	return s, func() {
+		if ts != nil {
+			ts.Close()
+		}
+	}
 }
 
-func testClientSuccessfulConnection(t *testing.T, svc *lambda.Lambda) {
-	TestDaemon.Reset()
-	ctx, root := BeginSegment(context.Background(), "Test")
+func testClientSuccessfulConnection(ctx context.Context, td *TestDaemon, t *testing.T, svc *lambda.Lambda) {
+	ctx, root := BeginSegment(ctx, "Test")
 	_, err := svc.ListFunctionsWithContext(ctx, &lambda.ListFunctionsInput{})
 	root.Close(nil)
 	assert.NoError(t, err)
 
-	s, e := TestDaemon.Recv()
-	assert.NoError(t, e)
+	seg, err := td.Recv()
+	if !assert.NoError(t, err) {
+		return
+	}
 
-	subseg := &Segment{}
-	assert.NotEmpty(t, s.Subsegments)
-	assert.NoError(t, json.Unmarshal(s.Subsegments[0], &subseg))
+	var subseg *Segment
+	if !assert.NoError(t, json.Unmarshal(seg.Subsegments[0], &subseg)) {
+		return
+	}
 	assert.False(t, subseg.Fault)
 	assert.NotEmpty(t, subseg.Subsegments)
 
@@ -144,19 +167,21 @@ func testClientSuccessfulConnection(t *testing.T, svc *lambda.Lambda) {
 	}
 }
 
-func testClientFailedConnection(t *testing.T, svc *lambda.Lambda) {
-	TestDaemon.Reset()
-	ctx, root := BeginSegment(context.Background(), "Test")
+func testClientFailedConnection(ctx context.Context, td *TestDaemon, t *testing.T, svc *lambda.Lambda) {
+	ctx, root := BeginSegment(ctx, "Test")
 	_, err := svc.ListFunctionsWithContext(ctx, &lambda.ListFunctionsInput{})
 	root.Close(nil)
 	assert.Error(t, err)
 
-	s, e := TestDaemon.Recv()
-	assert.NoError(t, e)
+	seg, err := td.Recv()
+	if !assert.NoError(t, err) {
+		return
+	}
 
-	subseg := &Segment{}
-	assert.NotEmpty(t, s.Subsegments)
-	assert.NoError(t, json.Unmarshal(s.Subsegments[0], &subseg))
+	var subseg *Segment
+	if !assert.NoError(t, json.Unmarshal(seg.Subsegments[0], &subseg)) {
+		return
+	}
 	assert.True(t, subseg.Fault)
 	// Should contain 'marshal' and 'attempt' subsegments only.
 	assert.Len(t, subseg.Subsegments, 2)
@@ -178,39 +203,31 @@ func testClientFailedConnection(t *testing.T, svc *lambda.Lambda) {
 	assert.NotEmpty(t, connectSubseg.Subsegments)
 }
 
-func testClientWithoutSegment(t *testing.T, svc *lambda.Lambda) {
-	Configure(Config{ContextMissingStrategy: &TestContextMissingStrategy{}})
-	defer ResetConfig()
-
-	ctx := context.Background()
+func testClientWithoutSegment(ctx context.Context, td *TestDaemon, t *testing.T, svc *lambda.Lambda) {
 	_, err := svc.ListFunctionsWithContext(ctx, &lambda.ListFunctionsInput{})
 	assert.NoError(t, err)
 }
 
-func testAWSDataRace(t *testing.T, svc *lambda.Lambda) {
-	Configure(Config{ContextMissingStrategy: &TestContextMissingStrategy{},DaemonAddr: "localhost:3000"})
-	defer ResetConfig()
-
-	ctx, cancel := context.WithCancel(context.Background())
-
+func testAWSDataRace(ctx context.Context, td *TestDaemon, t *testing.T, svc *lambda.Lambda) {
+	ctx, cancel := context.WithCancel(ctx)
+	defer cancel()
 	ctx, seg := BeginSegment(ctx, "TestSegment")
 
-	wg := sync.WaitGroup{}
-
+	var wg sync.WaitGroup
 	for i := 0; i < 5; i++ {
-		if i!=3 && i!=2{
+		if i != 3 && i != 2 {
 			wg.Add(1)
 		}
 		go func(i int) {
-			if i!=3 && i!=2{
-				time.Sleep(1)
+			if i != 3 && i != 2 {
+				time.Sleep(time.Nanosecond)
 				defer wg.Done()
 			}
 			_, seg := BeginSubsegment(ctx, "TestSubsegment1")
-			time.Sleep(1)
+			time.Sleep(time.Nanosecond)
 			seg.Close(nil)
 			svc.ListFunctionsWithContext(ctx, &lambda.ListFunctionsInput{})
-			if i== 3 || i==2{
+			if i == 3 || i == 2 {
 				cancel() // cancel context
 			}
 		}(i)
--- a/xray/capture_test.go
+++ b/xray/capture_test.go
@@ -20,64 +20,58 @@ import (
 )
 
 func TestSimpleCapture(t *testing.T) {
-	TestDaemon.Reset()
-	ctx, root := BeginSegment(context.Background(), "Test")
-	err := Capture(ctx, "TestService", func(ctx1 context.Context) error {
-		ctx = ctx1
-		defer root.Close(nil)
+	ctx, td := NewTestDaemon()
+	defer td.Close()
+
+	ctx, root := BeginSegment(ctx, "Test")
+	err := Capture(ctx, "TestService", func(context.Context) error {
+		root.Close(nil)
 		return nil
 	})
 	assert.NoError(t, err)
 
-	s, e := TestDaemon.Recv()
-	assert.NoError(t, e)
-	assert.Equal(t, "Test", s.Name)
-	assert.Equal(t, root.TraceID, s.TraceID)
-	assert.Equal(t, root.ID, s.ID)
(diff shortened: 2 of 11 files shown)


bhautikpip:
Hi @shogo82148,

I have investigated test case failures issue and it looks like we are not resetting a daemon because of that daemon sends incorrect data for some tests and this inconsistency happens after 10-15 retries. So I will create a PR which adds TestDaemon.Reset() before calling the TestDaemon.Recv() method. It solves the inconsistency of tests.

shogo82148:
`NewTestDaemon` overrides the daemon's address, but some clients to connect the default address `127.0.0.1:2000`.

> https://github.com/aws/aws-xray-sdk-go/pull/177/checks?check_run_id=392856538#step:6:93
>
> 2020-01-16T09:50:59Z [ERROR] write udp 127.0.0.1:33701->127.0.0.1:2000: write: connection refused

It means someone else changed the configure...
maybe the environment values?

bhautikpip:
I am not sure but tests are behaving inconsistent.

bhautikpip:
I have merged one PR today to fix resetting the daemon issue but looks like there is one more issue around TestDaemon.Recv() which fails the tests.

bhautikpip:
I am also getting the same error that you are getting in this tests. (https://github.com/aws/aws-xray-sdk-go/pull/177/checks?check_run_id=394369996#step:6:80).

shogo82148:
https://github.com/aws/aws-xray-sdk-go/pull/177/checks?check_run_id=394461889#step:6:100

race condition?

> 	Error:		Received unexpected error parse //��j�EɳS<�[�y�%2F@�E4�v��A�y�R��s: net/url: invalid userinfo

I'm working on SQL in https://github.com/aws/aws-xray-sdk-go/pull/169
Just comment out here.

shogo82148:
DATARACE https://github.com/aws/aws-xray-sdk-go/pull/177/checks?check_run_id=394461930#step:6:59

shogo82148:
Could you review and/or merge, please?

There are still some TODOs, but they are not in the scope of this pull request.
I'm going to work on other pull requests.

bhautikpip:
Hi @shogo82148,

Thanks for the changes. Just wanted to make sure that these changes solves the inconsistency of tests. would you mind doing an empty commit so that tests run again because I wanted to see if tests are consistent with these changes.

shogo82148:
no, i don't mind.
thank you.

daemoncfg/daemon_config_test.go
```diff
@@ -30,49 +29,49 @@ func TestGetDaemonEndpoints1(t *testing.T) { // default address set to udp and t
 	assert.Equal(t, dEndpt.TCPAddr, tcpEndpt)
 }
 
-func TestGetDaemonEndpoints2(t *testing.T) { // default address set to udp and tcp
-	udpAddr := "127.0.0.1:4000"
-	tcpAddr := "127.0.0.1:5000"
-	udpEndpt, _ := resolveUDPAddr(udpAddr)
-	tcpEndpt, _ := resolveTCPAddr(tcpAddr)
+// func TestGetDaemonEndpoints2(t *testing.T) { // default address set to udp and tcp
```
bhautikpip: Any reason to comment out functions TestGetDaemonEndpoints2, TestGetDaemonEndpointsFromEnv1, TestGetDaemonEndpointsFromEnv2 ?

xray/aws_test.go
```diff
@@ -20,7 +28,7 @@ func TestAWS(t *testing.T) {
 	// Runs a suite of tests against two different methods of registering
 	// handlers on an AWS client.
 
-	type test func(*testing.T, *lambda.Lambda)
+	type test func(context.Context, *TestDaemon, *testing.T, *lambda.Lambda)
```
bhautikpip: good change!

xray/aws_test.go
```diff
@@ -178,39 +203,34 @@ func testClientFailedConnection(t *testing.T, svc *lambda.Lambda) {
...
 }
 
-func testClientWithoutSegment(t *testing.T, svc *lambda.Lambda) {
-	Configure(Config{ContextMissingStrategy: &TestContextMissingStrategy{}})
-	defer ResetConfig()
-
-	ctx := context.Background()
-	_, err := svc.ListFunctionsWithContext(ctx, &lambda.ListFunctionsInput{})
-	assert.NoError(t, err)
+func testClientWithoutSegment(ctx context.Context, td *TestDaemon, t *testing.T, svc *lambda.Lambda) {
```
bhautikpip: Any particular issue with the function ? How can we fix this ?

xray/aws_test.go
```diff
@@ -178,39 +203,34 @@ func testClientFailedConnection(t *testing.T, svc *lambda.Lambda) {
...
+	// FIXME!!
+	// Configure(Config{ContextMissingStrategy: &TestContextMissingStrategy{}})
+	// defer ResetConfig()
+	// _, err := svc.ListFunctionsWithContext(ctx, &lambda.ListFunctionsInput{})
+	// assert.NoError(t, err)
 }
 
-func testAWSDataRace(t *testing.T, svc *lambda.Lambda) {
-	Configure(Config{ContextMissingStrategy: &TestContextMissingStrategy{},DaemonAddr: "localhost:3000"})
```
bhautikpip: Why we are removing configure() and Resetconfig() ?

xray/capture_test.go
```diff
@@ -20,91 +20,91 @@ import (
...
-	assert.Equal(t, root.ID, s.ID)
-	assert.Equal(t, root.StartTime, s.StartTime)
-	assert.Equal(t, root.EndTime, s.EndTime)
-	assert.NotNil(t, s.Subsegments)
-	subseg := &Segment{}
-	assert.NoError(t, json.Unmarshal(s.Subsegments[0], &subseg))
-	assert.Equal(t, "TestService", subseg.Name)
-}
-
-func TestCaptureAysnc(t *testing.T) {
```
bhautikpip: Why have we removed captureAsync test ?

xray/client_test.go
```diff
@@ -47,45 +63,53 @@ func TestRoundTripper(t *testing.T) {
...
+
+	ch := make(chan XRayHeaders, 1)
 	ts := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
-		headers = ParseHeadersForTest(r.Header)
-		b := []byte(`200 - Nothing to see`)
-		responseContentLength = len(b)
+		ch <- ParseHeadersForTest(r.Header)
 		w.WriteHeader(http.StatusOK)
-		w.Write(b)
+		w.Write([]byte(content))
```
bhautikpip: error return value of w.Write is not checked

xray/client_test.go
```diff
@@ -97,275 +121,301 @@ func TestRoundTrip(t *testing.T) {
...
-		assert.NoError(t, e)
+		_, err := td.Recv()
+		assert.NoError(t, err)
 	}
-	wg.Wait()
 }
 
-func TestRoundTripHttp2Datarace(t *testing.T) {
-	TestDaemon.Reset()
+func TestRoundTripReuseHTTP2Datarace(t *testing.T) {
```
bhautikpip: would you mind explaining change in this function ?

xray/config_test.go
```diff
@@ -89,46 +88,46 @@ func popEnv(env []string) {
 	}
 }
 
-func ResetConfig() {
```
bhautikpip: What's the reason for commenting out this function ?

xray/default_emitter_test.go
```diff
@@ -89,35 +89,35 @@ func TestStreamingSegmentsOnGrandchildNode(t *testing.T) {
 	assert.Equal(t, 23, len(packSegments(root, nil)))
 }
 
-func TestStreamingSegmentsTreeHasOnlyOneBranch(t *testing.T) {
```
bhautikpip: Why did we comment this out ?

xray/handler_test.go
```diff
@@ -27,65 +23,73 @@ func TestNewFixedSegmentName(t *testing.T) {
...
+	assert.Equal(t, http.StatusOK, seg.HTTP.Response.Status)
+	assert.Equal(t, "POST", seg.HTTP.Request.Method)
+	assert.Equal(t, ts.URL+"/", seg.HTTP.Request.URL)
+	assert.Equal(t, "127.0.0.1", seg.HTTP.Request.ClientIP)
+	assert.Equal(t, "UnitTest", seg.HTTP.Request.UserAgent)
+	assert.Equal(t, "TestVersion", seg.Service.Version)
 }
 
 func TestHandlerWithContextForNonRootHandler(t *testing.T) {
-	TestDaemon.Reset()
```
bhautikpip: Why did we add 3 functions in 1 function ?

xray/sql_test.go
```diff
@@ -8,254 +8,247 @@
...
-
-func (s *sqlTestSuite) TestDriverVersionPackage() {
-	s.mockDB("")
-	s.mockPSQL(nil)
-	s.connect()
-
-	s.Require().NoError(s.mock.ExpectationsWereMet())
-	//s.Equal("gopkg.in/DATA-DOG/go-sqlmock.v1", s.db.driverVersion)
-}
+// FIXME: @shogo82148
```
bhautikpip: Would you mind elaborating what type of fix this tests require ?

bhautikpip:
Hi @shogo82148,

I have reviewed this PR. Most of the questions related to commenting out few test cases and would you mind doing a few empty commit to trigger git tests and Travis build ? I am not sure if we can manually start running git tests ?

daemoncfg/daemon_config_test.go
shogo82148: added by https://github.com/aws/aws-xray-sdk-go/pull/177/commits/645b4bda883f350cabc08504f4a751e7e4c621e2

xray/aws_test.go
shogo82148: Configure and ResetConfig modify the environment values,
it effects the other tests which get the configure from the environment values.

see https://github.com/aws/aws-xray-sdk-go/pull/177#issuecomment-575416696

xray/aws_test.go
shogo82148: same as https://github.com/aws/aws-xray-sdk-go/pull/177/files#r368185353

xray/config_test.go
shogo82148: same reason as https://github.com/aws/aws-xray-sdk-go/pull/177#discussion_r368185353
and TestDefaultConfigureParameters doesn't touch the environment values.

xray/aws_test.go
shogo82148: oh, I have a misunderstanding.
Configure and ResetConfig don't modify the environment values.

But they still modify global configure. it may cause race condition.
I think we must not use them in tests.

xray/default_emitter_test.go
shogo82148: same reason as https://github.com/aws/aws-xray-sdk-go/pull/177#discussion_r368186749

xray/handler_test.go
shogo82148: I moved TestRootHandler and TestNonRootHandler to `xray/handler_go113_test.go`,
because I have no idea override the configure with `xray.Handler`.

easiest way is using `Configure`, but I can't because of https://github.com/aws/aws-xray-sdk-go/pull/177#discussion_r368186749

another way is using `Server.BaseContext` https://golang.org/pkg/net/http/#Server

```go
	ts.Config.BaseContext = func(_ net.Listener) context.Context {
		return ctx
	}
```

but it is available from Go 1.13, and minimum version of X-Ray SDK is Go 1.9.
So, I moved them to `xray/handler_go113_test.go`, and add build constraint ( `// +build go1.13` )

xray/client_test.go
shogo82148: it seems that `http2.ConfigureTransport(tr)` doesn't work we want.
we need to configure the server not only the client.

```go
	// configure http/2
	if err := http2.ConfigureServer(ts.Config, nil); !assert.NoError(t, err) {
		return
	}
	ts.TLS = ts.Config.TLSConfig
	ts.StartTLS()
```

oh.. Here is my mistake of copy.
It's from `TestRoundTripReuseTLSDatarace`

```go
	var wg sync.WaitGroup
	n := 100
	wg.Add(n)
	for i := 0; i < n; i++ {
		go func() {
			defer wg.Done()
			ctx, cancel := context.WithCancel(ctx)
			defer cancel()
			err := httpDoTest(ctx, client, http.MethodGet, ts.URL, nil)
			assert.NoError(t, err)
		}()
	}
	wg.Wait()
```

xray/capture_test.go
shogo82148: It's TestCapture A**ys**nc, not TestCaptureA**sy**nc.
and there is already TestCaptureA**sy**nc in the end of this file.

bhautikpip:
Looks like something is happening at this line (https://github.com/aws/aws-xray-sdk-go/pull/177/files#diff-05d404a7fd0c9af128e5597bba25efc1R423). There is one test case failed in every commit because of this.

shogo82148:
yeah... I'm trying to fix it, but it still fails...
Can I comment out and mark as TODOs?

bhautikpip:
I was testing this changes made by you to find out why tests are failing. I think if you comment out func TestRoundTripReuseDatarace(t *testing.T), func TestRoundTripReuseTLSDatarace(t *testing.T) and func TestRoundTripHttp2Datarace(t *testing.T) in client_test.go then tests will pass consistently and I have tested that. So my guess would be data race is happening inside these functions might want to look at that. I am not sure about what you have asked might want to consult with my team and get back to you if that works.

shogo82148:
I commented out these tests.

Currently some tests fail, but due to this pull request.
I issued that.
https://github.com/aws/aws-xray-sdk-go/pull/177/checks?check_run_id=397261225#step:6:55
https://github.com/aws/aws-xray-sdk-go/issues/180

bhautikpip:
I ran tests again. I think that error is related to the issue (https://github.com/aws/aws-xray-sdk-go/issues/180) you have opened and it's very rare still need to investigate on that but you will not see (https://github.com/aws/aws-xray-sdk-go/runs/396293215#step:6:75) this error after commenting out those functions in client_test.

shogo82148:
I misunderstood `-parallel n` option...
parallel execution is enabled only if we call `t. Parallel`
so touching the environment is safe.

bhautikpip:
Hi @shogo82148,

Do you think test cases are consistent now with this changes ?

shogo82148:
Yes, the are.

unfortunately, I give up fixing TestRoundTripReuseDatarace, TestRoundTripReuseTLSDatarace and TestRoundTripReuseHTTP2Datarace completely... 🙁
The aim of these test is detecting data race of http clients,
so there is no need to check that ALL X-Ray segments are received by the X-Ray daemon.
https://github.com/aws/aws-xray-sdk-go/pull/177/files#diff-05d404a7fd0c9af128e5597bba25efc1R340

bhautikpip:
Hi @shogo82148,

I think you are right. We might not want to check whether all the segments are received by the daemon. Do you think that is the reason causing test case failures ?

shogo82148:
yes.

bhautikpip:
Hi @shogo82148,

Are you going to work on SQL tests ? looks like there is some data race going on. Should we merge it then ?

shogo82148:
> Are you going to work on SQL tests ?

Yes, I am. But now, I'm waiting for the answer of https://github.com/aws/aws-xray-sdk-go/pull/169

> Should I consider merging this PR ?

Yes, please.

bhautikpip:
Okay. I will expedite the review process for (https://github.com/aws/aws-xray-sdk-go/pull/169). It looks like a good support. Also, would you mind commenting out this line in reservoir_test.go (https://github.com/aws/aws-xray-sdk-go/blob/master/strategy/sampling/reservoir_test.go#L46) and add a line FIXME or something and remove go.mod changes from the commit (https://github.com/aws/aws-xray-sdk-go/pull/177/files#diff-37aff102a57d3d7b797f152915a6dc16R3) ?

bhautikpip:
I was also talking about this too. (https://github.com/aws/aws-xray-sdk-go/pull/177/files#diff-2edb5191aa62c1d54488138489fe9745R46)

shogo82148:
done!

xray/util_test.go
```diff
@@ -9,93 +9,120 @@
...
+		case <-td.ctx.Done():
+			return
+		}
 	}
 }
 
-func (td *Testdaemon) Recv() (*Segment, error) {
-	ctx, cancel := context.WithTimeout(context.Background(), 500*time.Millisecond)
+func (td *TestDaemon) Recv() (*Segment, error) {
+	ctx, cancel := context.WithTimeout(td.ctx, 5*time.Second)
```
willarmiros: Why increase the timeout length from 500 milliseconds to 5 seconds?

xray/sql_test.go
```diff
@@ -8,254 +8,247 @@
...
-	s.mockDB("")
-	s.mockPSQL(nil)
-	s.connect()
-
-	s.Require().NoError(s.mock.ExpectationsWereMet())
-	//s.Equal("gopkg.in/DATA-DOG/go-sqlmock.v1", s.db.driverVersion)
-}
+// FIXME: @shogo82148 remove race condition.
+// see https://github.com/aws/aws-xray-sdk-go/pull/177#issuecomment-575438315
+// func TestSQL(t *testing.T) {
```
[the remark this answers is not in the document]
shogo82148: OK

xray/util_test.go
shogo82148: I did it for trying to fix TestRoundTripReuseDatara, TestRoundTripReuseTLSDatarace, TestRoundTripHttp2Datarace.
but it doesn't work I expected. I'll revert it.

PR #178, opened 2020-01-16
fixed inconsistent test failures and sql test failure

bhautikpip:
*Issue #, if available:*

*Description of changes:*

1. Major Bugfix: Inconsistent test failures because of daemon sending incorrect data after 10-15 attempts
2. Fixed sql test cases by changing ports to numeric (recent change in Go)
3. Fixed test.yml (windows 9 and 10 environment set up)

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -29,11 +29,14 @@ jobs:
       - name: Set up GOPATH
         run: |
           go version
-          echo "::set-env name=GOPATH::$GITHUB_WORKSPACE"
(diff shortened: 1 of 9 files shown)


chanchiem:
Is there a reason why we can't use setup/teardown instead of manually invoking it in each of the tests? I think cleaning up states should almost always be done between each unit test anyway.

bhautikpip:
I think there are multiple ways we can do this but just wanted to do sanity check before calling TestDaemon.Recv(). what do you mean by set up ?

PR #183, opened 2020-01-22
avoid multiple initialization of rand

shogo82148:
*Issue #, if available:*

It looks that `math/rand` is initialized two times.

- https://github.com/aws/aws-xray-sdk-go/blob/dc1201085c0b1e64259ef8f7cb48954691212db9/utils/rand.go#L17
- https://github.com/aws/aws-xray-sdk-go/blob/dc1201085c0b1e64259ef8f7cb48954691212db9/utils/timer.go#L17

`math/rand.Seed` changes global state of users' program, so any SDK should not call this.

*Description of changes:*

I implemented a random value generator only used by the X-Ray SDK.


bhautikpip:
Hi @shogo82148,

Thanks for opening up this PR. changes looks good! I might be able to merge it once we resolve SQL test case failure issue.

utils/rand.go
```diff
@@ -9,12 +9,72 @@
...
+func newSeed() int64 {
+	var seed int64
+	if err := binary.Read(crand.Reader, binary.BigEndian, &seed); err != nil {
+		// fallback to timestamp
+		seed = time.Now().UnixNano()
+	}
+	return seed
+}
+
+func newGlobalRand() *rand.Rand {
```
bhautikpip: Can we add tests for this function which covers all the combination ?

bhautikpip:
Hi @shogo82148 ,

would you mind resolving the conflict and adding few tests around the function ?

shogo82148:
resolving the conflict is done!

Issue #184, opened 2020-01-22
bug: Logging outputs to many lines


sthulb:
The built in logging reports a single JSON object over multiple lines. It should output this over a single line.

```
timestamp="2020-01-22T10:49:22.4914885Z" severity=info message="2020-01-22T10:49:22Z [DEBUG] {"
timestamp="2020-01-22T10:49:22.4915148Z" severity=info message=" \"trace_id\": \"1-dc99d00f-c079a84d433534434534ef0d\","
timestamp="2020-01-22T10:49:22.4915396Z" severity=info message=" \"id\": \"01694e90f21e74b0\","
timestamp="2020-01-22T10:49:22.4915597Z" severity=info message=" \"name\": \"service\","
timestamp="2020-01-22T10:49:22.491584Z" severity=info message=" \"start_time\": 1579690162.4909065,"
timestamp="2020-01-22T10:49:22.4916041Z" severity=info message=" \"end_time\": 1579690162.4909775,"
timestamp="2020-01-22T10:49:22.4916287Z" severity=info message=" \"parent_id\": \"91ed514f1e5c03b2\","
timestamp="2020-01-22T10:49:22.4916529Z" severity=info message=" \"type\": \"subsegment\","
timestamp="2020-01-22T10:49:22.4916767Z" severity=info message=" \"annotations\": {"
timestamp="2020-01-22T10:49:22.4917006Z" severity=info message="  \"service\": \"ServiceName\""
timestamp="2020-01-22T10:49:22.491729Z" severity=info message=" }"
timestamp="2020-01-22T10:49:22.4917494Z" severity=info message="}"
```

bhautikpip:
Merged this PR (https://github.com/aws/aws-xray-sdk-go/pull/185). So this should be solved. So, I am closing this issue feel free to reopen this if you see this again.

PR #185, opened 2020-01-22
fix: remove json.Indent from segment debug logging

sthulb:
*Issue #, if available:*
#184 
*Description of changes:*
Removes `json.Ident` from segment debug log. By using `json.Ident` the SDK will break logging tools that want to make use of this data. For example, this will break CloudWatch Logs Insights since it won't be able to inspect the JSON object.


bhautikpip:
Hi @sthulb,

Thanks for the contribution. I have reviewed the PR. I might be able to merge this once we resolve the SQL test cases failure issue if that's okay.

PR #187, opened 2020-01-29
Begin segment sampling

bhautikpip:
*Issue #, if available:*

*Description of changes:*

1. Modified BeginSegment API to add localized sampling in the case where customer use this API without instrumenting their application
2. If customer pass true then it will do localized sampling and if they pass falls then API will do all segment sampling
3. This change is consistent with other internal calls


xray/segment.go
```diff
@@ -53,7 +54,7 @@ func BeginFacadeSegment(ctx context.Context, name string, h *header.Header) (con
 }
 
 // BeginSegment creates a Segment for a given name and context.
-func BeginSegment(ctx context.Context, name string) (context.Context, *Segment) {
+func BeginSegment(ctx context.Context, name string, sample bool) (context.Context, *Segment) {
```
willarmiros: As discussed, we do not want to break the external API here. We should probably add a new API, e.g. `BeginSegmentWithSampling`, that has these parameters.

bhautikpip: Right. Makes Sense!

xray/segment.go
```diff
@@ -206,8 +264,8 @@ func BeginSubsegment(ctx context.Context, name string) (context.Context, *Segmen
 }
 
 // NewSegmentFromHeader creates a segment for downstream call and add information to the segment that gets from HTTP header.
-func NewSegmentFromHeader(ctx context.Context, name string, h *header.Header) (context.Context, *Segment) {
-	con, seg := BeginSegment(ctx, name)
+func NewSegmentFromHeader(ctx context.Context, name string, r *http.Request, h *header.Header) (context.Context, *Segment) {
```
willarmiros: In line with above comments, would not need `*http.Request` parameter anymore.

xray/handler.go
```diff
@@ -138,19 +134,6 @@ func httpTrace(seg *Segment, h http.Handler, w http.ResponseWriter, r *http.Requ
 	respHeader.WriteString("Root=")
 	respHeader.WriteString(seg.TraceID)
 
-	if traceHeader.SamplingDecision != header.Sampled && traceHeader.SamplingDecision != header.NotSampled {
```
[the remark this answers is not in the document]
willarmiros: Resolved since this change is necessary to accommodate dummy segment support.

xray/segment.go
willarmiros: Resolved - this is fine due to dummy seg support.

xray/segment.go
```diff
@@ -79,6 +87,56 @@ func BeginSegment(ctx context.Context, name string) (context.Context, *Segment)
 	return context.WithValue(ctx, ContextKey, seg), seg
 }
 
+func BeginSegmentWithSamplingDecision(ctx context.Context, name string, r *http.Request, traceHeader *header.Header) (context.Context, *Segment) {
```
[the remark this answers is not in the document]
willarmiros: Update: The first bullet point in this comment can now be ignored, however I do think the name of the function should be updated to better reflect the parameters it takes. Something like `BeginSegmentWithHeader`. 

The second two bullets still apply.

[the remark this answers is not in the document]
bhautikpip: Changed the name to BeginSegmentWithSampling as we are performing sampling while creating segment. Moreover, unit tests to validate this function will be the same as verifying seg.Sampled field. We have already tests for it (https://github.com/aws/aws-xray-sdk-go/blob/master/strategy/sampling/centralized_test.go). In this function I am just calling those methods in order to populate seg.Sampled.

xray/segment.go
```diff
@@ -68,6 +75,40 @@ func BeginSegment(ctx context.Context, name string) (context.Context, *Segment)
 		seg.GetService().Version = seg.ParentSegment.GetConfiguration().ServiceVersion
 	}
 
+	// Sampling strategy fallbacks to default sampling in the case of directly using BeginSegment API without any instrumentation
+	if r == nil && traceHeader == nil {
```
willarmiros: I think this condition can be a logical or (`||`). If, for example, a customer calls this API and leaves one of the parameters nil we still can't infer a sampling decision and thus should fall back to the default sampling strategy.

xray/segment.go
```diff
@@ -68,6 +75,40 @@ func BeginSegment(ctx context.Context, name string) (context.Context, *Segment)
...
+		sd := seg.ParentSegment.GetConfiguration().SamplingStrategy.ShouldTrace(&sampling.Request{})
+		seg.Sampled = sd.Sample
+		logger.Debugf("SamplingStrategy decided: %t", seg.Sampled)
+		seg.AddRuleName(sd)
+	}
+
+	// Sampling strategy for http calls
+	if r != nil && traceHeader != nil {
```
willarmiros: Making the above change means this can just be an `else` condition, which would be more concise.

xray/segment.go
bhautikpip: good catch!

PR #193, opened 2020-02-12
Global sanitizing support

bhautikpip:
*Issue #, if available:*

*Description of changes:*

Added global sanitizing strategy for HTTP, SQL, AWS and metadata. Completed SQL sanitizer still working on HTTP, AWS and metadata. Created this PR to ask feedback about design of global sanitizing strategy. 


strategy/sanitizing/sanitizer.go
```diff
@@ -0,0 +1,18 @@
...
+//     http://aws.amazon.com/apache2.0/
+//
+// or in the "license" file accompanying this file. This file is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the specific language governing permissions and limitations under the License.
+
+package sanitizing
+
+// Strategy provides an interface for
+// implementing different sanitizers
+type Strategy interface {
+	SqlSanitizer(san string) string
```
shogo82148: method name should be verb. (e.g. Sanitize)
"Sql" should be "SQL", see https://github.com/golang/go/wiki/CodeReviewComments#initialisms

strategy/sanitizing/sanitizer.go
```diff
@@ -0,0 +1,18 @@
...
+	HttpSanitizer(san string) string
```
shogo82148: "Http" should be "HTTP" https://github.com/golang/go/wiki/CodeReviewComments#initialisms

xray/config.go
```diff
@@ -118,6 +126,7 @@ type Config struct {
 	StreamingStrategy           StreamingStrategy
 	ExceptionFormattingStrategy exception.FormattingStrategy
 	ContextMissingStrategy      ctxmissing.Strategy
+	sanitizingStrategy          sanitizing.Strategy
```
shogo82148:
```suggestion
	SanitizingStrategy          sanitizing.Strategy
```

strategy/sanitizing/sanitizer.go
```diff
@@ -0,0 +1,18 @@
...
+package sanitizing
+
+// Strategy provides an interface for
+// implementing different sanitizers
+type Strategy interface {
+	SQLSanitizer(san string) string
+	HTTPSanitizer(san string) string
+	AWSSanitizer(san string) string
+	MetadataSanitizer(san string) string
+}
```
shengxil: Missing Annotation sanitizer. Also I suggest to have a single method that takes the type as parameter, i.e.
`sanitize(type SanitizingType, key string, value string)`
This will be more extensible as when we add more supported fields it won't break the interface. The method should take a `key` as well because there are many keys under SQL, e.g. host, query, version, etc.

strategy/sanitizing/sanitizing_strategy.go
```diff
@@ -0,0 +1,55 @@
...
+
+		san = buf.String()
+	}
+
+	return san
+}
+
+func (ds *DefaultSanitizer) HTTPSanitizer(san string) string {
+	// Code for HTTP Sanitizer
+	return ""
```
shengxil: It should return the `san` rather than an empty string

strategy/sanitizing/sanitizing_strategy.go
```diff
@@ -0,0 +1,55 @@
...
+		buf.WriteString(str2)
+
+		san = buf.String()
+	}
+
+	return san
+}
+
+func (ds *DefaultSanitizer) HTTPSanitizer(san string) string {
+	// Code for HTTP Sanitizer
```
shengxil: Redundant comment

strategy/sanitizing/sanitizer.go
bhautikpip: good suggestion!

strategy/sanitizing/sanitizing_strategy.go
```diff
@@ -0,0 +1,69 @@
...
+type Sanitizer struct{}
+
+// NewDefaultSanitizingStrategy initializes
+// an instance of DefaultSanitizer.
+func NewSanitizingStrategy() (*Sanitizer, error) {
+	return &Sanitizer{}, nil
+}
+
+// DefaultSanitizer provides basic sanitizing implementation
+func (ds *Sanitizer) DefaultSanitizer(san string, value string) string {
```
shengxil: Shall we rename the "san" to "type" or "key" that make better sense?

strategy/sanitizing/sanitizing_strategy.go
```diff
@@ -0,0 +1,69 @@
...
+
+	case "AWS":
+		// Code for AWS Sanitizer
+		return value
+
+	case "Metadata":
+		// Code for Metadata Sanitizer
+		return value
+
+	case "Annotation":
```
shengxil: Would it be enough for just passing a value for annotation? I'd expect customer want to sanitize differently by the annotation key. Same as HTTP/AWS/Metadata.

xray/segment.go
```diff
@@ -52,6 +52,19 @@ func BeginFacadeSegment(ctx context.Context, name string, h *header.Header) (con
 	return context.WithValue(ctx, ContextKey, seg), seg
 }
 
+// BeginDefaultSanitizing returns sanitized string for a given san(SQL, HTTP, AWS, Metadata, Annotations).
+func BeginDefaultSanitizing(san string, value string) string {
```
shengxil: Does customer have to explicitly call `BeginDefaultSanitizing` before he adds the data? Can we do this for him?

PR #194, opened 2020-02-13
Dummy flag to reduce operations on non sampled traces

bhautikpip:
*Issue #, if available:*

*Description of changes:*

Added support for  creating DummySegment and DummySubSegments. 


xray/segment.go
```diff
@@ -54,6 +54,25 @@ func BeginFacadeSegment(ctx context.Context, name string, h *header.Header) (con
...
 }
 
+// Begin DummySegment creates a segment in the case of no sampling to reduce memory footprint
+func BeginDummySegment(ctx context.Context, name string) (context.Context, *Segment) {
+	dummySeg := &Segment{parent: nil}
+	dummySeg.ParentSegment = dummySeg
+	logger.Debugf("Beginning segment named %s", name)
+
+	cfg := GetRecorder(ctx)
+	dummySeg.assignConfiguration(cfg)
```
srprash: Is it necessary to pass the recorder in configuration for a DummySegment? The `assignConfiguration` method does a bunch of comparisons which should be avoided if we don't really need the configuration.

xray/segment.go
```diff
@@ -54,6 +54,25 @@ func BeginFacadeSegment(ctx context.Context, name string, h *header.Header) (con
...
+
+	dummySeg.Lock()
+	defer dummySeg.Unlock()
+
+	dummySeg.Name = name
+	dummySeg.TraceID = "dummy segment"
```
srprash: Can be an empty string or something more meaningful.

xray/segment.go
```diff
@@ -116,7 +135,11 @@ func BeginSegmentWithSampling(ctx context.Context, name string, r *http.Request,
 		}()
 	}
 
-	return context.WithValue(ctx, ContextKey, seg), seg
+	if seg.Sampled {
+		return context.WithValue(ctx, ContextKey, seg), seg
+	}
+
+	return BeginDummySegment(ctx, "Dummy Seg")
```
srprash: Better to use full name as "DummySegment"

xray/segment.go
```diff
@@ -198,6 +221,33 @@ func (seg *Segment) assignConfiguration(cfg *Config) {
...
+
+	dummySubSeg := &Segment{parent: parent}
+	logger.Debugf("Beginning subsegment named %s", name)
+
+	dummySubSeg.Lock()
+	defer dummySubSeg.Unlock()
+
+	dummySubSeg.ParentSegment = parent.ParentSegment
+	dummySubSeg.Name = name
+	dummySubSeg.TraceID = "dummy subsegment"
```
srprash: Same here

xray/segment.go
```diff
@@ -242,7 +292,11 @@ func BeginSubsegment(ctx context.Context, name string) (context.Context, *Segmen
 	seg.StartTime = float64(time.Now().UnixNano()) / float64(time.Second)
 	seg.InProgress = true
 
-	return context.WithValue(ctx, ContextKey, seg), seg
+	if seg.ParentSegment.Sampled {
+		return context.WithValue(ctx, ContextKey, seg), seg
+	}
+
+	return BeginDummySegment(ctx, "Dummy SubSeg")
```
srprash: Better to use full name as "DummySubsegment"

xray/segment.go
```diff
@@ -54,6 +54,25 @@ func BeginFacadeSegment(ctx context.Context, name string, h *header.Header) (con
 	return context.WithValue(ctx, ContextKey, seg), seg
 }
 
+// Begin DummySegment creates a segment in the case of no sampling to reduce memory footprint
+func BeginDummySegment(ctx context.Context, name string) (context.Context, *Segment) {
+	dummySeg := &Segment{parent: nil}
```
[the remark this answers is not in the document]
bhautikpip: I think we are making sure that we get nil parent and not relying on initialization.

xray/segment.go
```diff
@@ -54,6 +54,25 @@ func BeginFacadeSegment(ctx context.Context, name string, h *header.Header) (con
...
+	dummySeg.ParentSegment = dummySeg
+	logger.Debugf("Beginning segment named %s", name)
```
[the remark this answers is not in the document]
bhautikpip: Good pick

xray/segment.go
bhautikpip: right.

xray/segment.go
```diff
@@ -198,6 +221,33 @@ func (seg *Segment) assignConfiguration(cfg *Config) {
...
+	if parent == nil {
+		cfg := GetRecorder(ctx)
+		failedMessage := fmt.Sprintf("failed to begin subsegment named '%v': segment cannot be found.", name)
+		if cfg != nil && cfg.ContextMissingStrategy != nil {
+			cfg.ContextMissingStrategy.ContextMissing(failedMessage)
+		} else {
+			globalCfg.ContextMissingStrategy().ContextMissing(failedMessage)
+		}
+		return ctx, nil
+	}
```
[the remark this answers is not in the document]
bhautikpip: I think this is necessary because the context passed in dummy subsegment and normal subsegment will not be the same. So this context will return us dummy segment.

xray/segment.go
```diff
@@ -116,7 +135,11 @@ func BeginSegmentWithSampling(ctx context.Context, name string, r *http.Request,
 		}()
 	}
 
-	return context.WithValue(ctx, ContextKey, seg), seg
+	if seg.Sampled {
+		return context.WithValue(ctx, ContextKey, seg), seg
+	}
+
+	return BeginDummySegment(ctx, "DummySegment")
```
yogiraj07: If the Segment is dummy, adding annotation and metadata should be no-op to save the memory footprint.

xray/segment.go
```diff
@@ -263,6 +270,12 @@ func NewSegmentFromHeader(ctx context.Context, name string, r *http.Request, h *
 // Close a segment.
 func (seg *Segment) Close(err error) {
 	seg.Lock()
+
+	// If segment is dummy we return
+	if seg.Dummy {
```
shengxil, line 275: Should this be placed before `seg.Lock()`? What happens if segment left locked because it's a dummy?

xray/segment.go
```diff
@@ -449,15 +477,20 @@ func (seg *Segment) beforeEmitSubsegment(s *Segment) {
 
 // AddAnnotation allows adding an annotation to the segment.
 func (seg *Segment) AddAnnotation(key string, value interface{}) error {
+	seg.Lock()
```
shengxil, line 480: why do we lock and unlock before checking dummy?

xray/segment.go
```diff
@@ -240,6 +244,9 @@ func BeginSubsegment(ctx context.Context, name string) (context.Context, *Segmen
 	seg.StartTime = float64(time.Now().UnixNano()) / float64(time.Second)
 	seg.InProgress = true
 
+	// check whether segment is dummy or not based on sampling decision
+	seg.isDummy()
```
shengxil, line 248: can we fail faster by putting this immediately after `seg.ParentSegment = parent.ParentSegment`? We don't bother  keeping the footprint in the parent segment as we won't send it out anyway.

xray/segment.go
bhautikpip: because we don't want any data race on segment struct. so it's a good idea to lock the struct before accessing it.

xray/segment.go
bhautikpip: Yeah. we can put this right after seg.ParentSegment = parent.ParentSegment.

xray/segment.go
bhautikpip: Yes. This is a good point. We should unlock the segment before returning in the case where segment is dummy.

xray/segment.go
```diff
@@ -409,6 +428,16 @@ func (seg *Segment) root() *Segment {
 	return seg.parent.root()
 }
 
+// check the segment is dummy segment or not
+func (seg *Segment) isDummy() *Segment {
```
shengxil, line 432: This method seems not necessary. It is only called at 2 places when begin a segment or subsegment. Customer doesn't need to call it as well. I'd suggest replacing the caller code with
 ```
if !seg.ParentSegment.Sampled 
   seg.Dummy = true
```
BTW The method name is confusing. `isXYZ()` should returns a boolean. Now the return type is never used (see the github warn below)

Comment (bhautikpip) [reply xray/segment.go]: good suggestion!

File: xray/segment.go [hunk 14]
```diff
@@ -283,6 +301,11 @@ func (seg *Segment) CloseAndStream(err error) {
 	seg.Lock()
 	defer seg.Unlock()
 
+	// If segment is dummy we return
+	if seg.Dummy {
```
shengxil, line 305: At creating segment we dealt with the segment children membership no matter dummy or not. If we return early here, would skipping the following code cause mem leak? i.e. it will always be "in progress" and the parent always keep the reference of it

shengxil: it should be semantic

PR #219, opened 2020-04-12
Disabling sdk feature

bhautikpip:
*Issue #, if available:*

*Description of changes:*

Added environment variable support to disable SDK feature.


xray/segment.go
```diff
@@ -61,6 +62,13 @@ func BeginSegment(ctx context.Context, name string) (context.Context, *Segment)
...
 
 func BeginSegmentWithSampling(ctx context.Context, name string, r *http.Request, traceHeader *header.Header) (context.Context, *Segment) {
+	// If SDK is disabled then return with an empty segment
+	disableKey := os.Getenv("AWS_XRAY_SDK_DISABLED")
+	if strings.ToLower(disableKey) == "true" {
+		seg := &Segment{}
+		return context.WithValue(ctx, ContextKey, seg), seg
+	}
+
 	seg := basicSegment(name, nil)
```
willarmiros, line 72: You should either construct a basic segment or at least populate the `parentSegment` of the new segment before returning it IMO. There are several methods that assume the fields populated in `basicSegment` are non-nil, e.g. `DownstreamHeader()` in `segment_model.go`. If you return an empty segment, the `parentSegment` field will be `nil` and the method will throw a nil dereference error.

xray/segment.go
```diff
@@ -204,6 +212,13 @@ func (seg *Segment) assignConfiguration(cfg *Config) {
 
 // BeginSubsegment creates a subsegment for a given name and context.
 func BeginSubsegment(ctx context.Context, name string) (context.Context, *Segment) {
+	// If SDK is disabled then return with an empty segment
+	disableKey := os.Getenv("AWS_XRAY_SDK_DISABLED")
+	if strings.ToLower(disableKey) == "true" {
+		seg := &Segment{}
+		return context.WithValue(ctx, ContextKey, seg), seg
+	}
+
```
willarmiros, line 221: Same here, I think it at least needs `parentSegment` field populated.

dcu:
for anyone interested in testing this change, add this line to your go.mod:

```go
replace github.com/aws/aws-xray-sdk-go => github.com/bhautikpip/aws-xray-sdk-go Disable-SDK
```

and then run:

```
go mod tidy
```

xray/segment.go
bhautikpip: That's a good point. After scanning the code base only DownstreamHeader() method can cause the nil pointer exception because customer might call that method directly otherwise other APIs are used by SDK internally. Also, if we don't populate segment fields then we can avoid unnecessary operations in the case of beginsubsegment. In this draft I have added disableKey logic in methods which customer can directly call to avoid throwing panic.

willarmiros: Fair enough, this looks good to me!

xray/segment.go
```diff
@@ -61,6 +62,13 @@ func BeginSegment(ctx context.Context, name string) (context.Context, *Segment)
 }
 
 func BeginSegmentWithSampling(ctx context.Context, name string, r *http.Request, traceHeader *header.Header) (context.Context, *Segment) {
+	// If SDK is disabled then return with an empty segment
+	disableKey := os.Getenv("AWS_XRAY_SDK_DISABLED")
```
anuraaga, line 66: Does it make sense to put this in `config.go`? If not, at least we should extract a function for this instead of copying it everywhere.

bhautikpip: That's a fair point. We can extract a function instead of writing this logic everywhere.

xray/segment.go
```diff
@@ -61,6 +62,13 @@ func BeginSegment(ctx context.Context, name string) (context.Context, *Segment)
 }
 
 func BeginSegmentWithSampling(ctx context.Context, name string, r *http.Request, traceHeader *header.Header) (context.Context, *Segment) {
+	// If SDK is disabled then return with an empty segment
+	sdkDisable := Disabled()
```
anuraaga, line 66: Is there a reason to have this local variable instead of just `if Disabled()`?

xray/segment.go
```diff
@@ -61,6 +62,13 @@ func BeginSegment(ctx context.Context, name string) (context.Context, *Segment)
...
+	if sdkDisable {
+		seg := &Segment{}
```
anuraaga, line 68: As an aside, it would be less fragile if we returned a no-op recorder when the SDK is disabled (wouldn't be flippable at runtime but I don't think that's needed). A bigger change so can think about it for later.

xray/segment.go
```diff
@@ -271,8 +286,23 @@ func NewSegmentFromHeader(ctx context.Context, name string, r *http.Request, h *
 	return con, seg
 }
 
+// Check if SDK is disabled
+func Disabled() bool {
```
anuraaga, line 290: Especially if removing the local variable, probably good to name this `func SdkDisabled()`

xray/segment.go
bhautikpip: Makse sense! will modify it

xray/segment.go
bhautikpip: good point!

xray/segment.go
bhautikpip: right.

Issue #110, opened 2019-05-22
xray handler for grpc 


rvdwijngaard:
Do you have any plans for adding xray handler for grpc servers / clients?

ss2305:
Hi Ron,

Currently we do not have support for grpc but we will add it into our backlogs. Can you share examples of use cases you are trying to tackle? 
Meanwhile we do welcome PRs and would be happy to review. 

Best Regards
Srikant

shubharao:
App Mesh customers are asking for this: https://github.com/aws/aws-app-mesh-roadmap/issues/256

srprash:
Hi @rvdwijngaard 
We have this feature request added to our backlog. We would try to pull it in to work on as soon as possible. 
Thanks for your patience and please stay tuned!

JustinEdwardKennedy:
Created https://github.com/vendrive/xray-grpc for temporary use while https://github.com/aws/aws-xray-sdk-go/pull/264 makes its way through.

bhautikpip:
Instrumentation unary gRPC support is merged and released in `v1.5.0` release so closing out this issue.

PR #221, opened 2020-04-18
Started unary grpc support

stijndehaes:
*Issue #, if available:* https://github.com/aws/aws-xray-sdk-go/issues/110

*Description of changes:*
First stab at grpc support. Kept the scope for now to unary grpc support.
Used https://github.com/grpc-ecosystem/go-grpc-middleware as a basis for starting with this. 

Added simple documentation on how to use it.
This is WIP since I wanted to get the ball rolling on how to continue with this. This basic support already works. Not sure how to best improve this, so I am hoping from some input of the maintainers.


stijndehaes:
Example trace for the PR with aws app mesh in between.
The errors are because I made the server fail 50% of the time to test this behavior. Client reports less errors because aws app mesh has a retry policy.

image

xray/grpc_client.go
```diff
@@ -0,0 +1,17 @@
+package xray
+
+import (
+	"context"
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/metadata"
+)
+
+func UnaryClientInterceptor(name string) grpc.UnaryClientInterceptor {
```
bhautikpip, line 9: How do we plan to propagate context ? We might need to pass ctx with the name argument.

xray/grpc_server.go
```diff
@@ -0,0 +1,30 @@
...
+	"google.golang.org/grpc/status"
+)
+
+func UnaryServerInterceptor(name string) grpc.UnaryServerInterceptor {
+	return func(ctx context.Context, req interface{}, info *grpc.UnaryServerInfo, handler grpc.UnaryHandler) (interface{}, error) {
+		md, ok := metadata.FromIncomingContext(ctx)
+		if !ok {
+			return nil, status.Errorf(codes.InvalidArgument, "missing metadata")
+		}
+		traceId := md.Get(TraceIDHeaderKey)
```
bhautikpip, line 18: Can we follow review dog comments for naming variables ?

xray/grpc_server.go
```diff
@@ -0,0 +1,30 @@
...
+			traceIdHeader = traceId[0]
+		}
+		theader := header.FromString(traceIdHeader)
+		ctx, seg := NewSegmentFromHeader(ctx, name, nil, theader)
+		_ = seg.AddMetadata("method", info.FullMethod)
+		m, err := handler(ctx, req)
+		seg.Close(err)
+		return m, err
+	}
+}
```
bhautikpip, line 30: Can we add tests for gRPC server as well as client example ? So that we might be able to understand better which information should populate in the trace and then we can change accordingly

xray/grpc_client.go
vshkola: Hi,

Yeah, I agree that would be better to propagate existing context instead of creating a dummy segment, I have the following implementation for my project:

`func addTraceData(ctx context.Context, req interface{}) (context.Context, error) {
        // should contains traceID and segment in context
	seg := xray.GetSegment(ctx)
	if seg == nil || seg.TraceID == "" {
		return ctx, fmt.Errorf("No xray segment found")
	}
	subCtx, subSeg := xray.BeginSubsegment(ctx, "grpc-client")
	defer subSeg.Close(nil)

	md, ok := metadata.FromOutgoingContext(subCtx)
	if !ok {
		md = metadata.MD{}
	}
        // we can iterate over all values from the context and pass it too
	md.Set("traceID", seg.TraceID)
	md.Set("parentID", seg.ID)

	outCtx := metadata.NewOutgoingContext(subCtx, md)
	xray.AddMetadata(subCtx, "Request", req)

	return outCtx, nil
}`

and then use it in Interceptor before invoker.

But in order to trace all the info, would be better either to implement something similar as https://github.com/grpc-ecosystem/grpc-opentracing or enable support for OpenTracing in Xray.

stijndehaes:
Hi all I am going to close this PR as I don't find the time to finish this.

vshkola:
Hi @bhautikpip, I would like then to add my implementation in another PR, but before I will do that, do you have any plans to add OpenTracing format support to AWS X-Ray?

Issue #222, opened 2020-04-20
AWS requests hangs after a few iterations.


skaggmannen:
I have written a small test app to try and reproduce issues we saw in our X-Ray enabled Lambdas when upgrading from 1.0.0-rc.14 to 1.0.0. 

Basically it seems that AWS requests hangs after 2-3 requests being made from the same Lambda instance causing the Lambdas to timeout. Exactly the same thing happens when running the following app:

```go
package main

import (
	"bytes"
	"encoding/json"
	"log"
	"net/http"
	"net/http/httptest"
	"os"

	"github.com/aws/aws-sdk-go/aws/session"
	"github.com/aws/aws-sdk-go/service/dynamodb"
	"github.com/aws/aws-xray-sdk-go/xray"
	"github.com/aws/aws-xray-sdk-go/xraylog"
)

func main() {
	xray.SetLogger(xraylog.NewDefaultLogger(os.Stderr, xraylog.LogLevelDebug))

	sess := session.Must(session.NewSession())
	xray.AWSSession(sess)

	db := dynamodb.New(sess)

	var handler http.HandlerFunc = func(rsp http.ResponseWriter, req *http.Request) {
		log.Println("Listing tables...")
		output, err := db.ListTablesWithContext(req.Context(), &dynamodb.ListTablesInput{})
		if err != nil {
			panic(err)
		}

		rsp.WriteHeader(http.StatusOK)
		err = json.NewEncoder(rsp).Encode(output.TableNames)
		if err != nil {
			panic(err)
		}

		log.Println("Done!")
	}

	wrapped := xray.Handler(xray.NewFixedSegmentNamer("xray-test-app"), handler)

	for i := 0; i < 100; i++ {
		rsp := httptest.NewRecorder()
		req, err := http.NewRequest(http.MethodGet, "https://dummy/url", &bytes.Buffer{})
		if err != nil {
			panic(err)
		}

		wrapped.ServeHTTP(rsp, req)
		if rsp.Code != http.StatusOK {
			panic("expected ok")
		}
	}
}

```

Is this a bug introduced in 1.0.0 or am I doing something terribly wrong?

bhautikpip:
Hi @skaggmannen ,

[...]

PR #223, opened 2020-04-20
Fixed deadlock issue for non-reported segments.

skaggmannen:
*Issue #, if available: https://github.com/aws/aws-xray-sdk-go/issues/222*

*Description of changes:*
A recent introduction if the `Dummy` property caused a function on `Segment` to return before the lock had been released.

### DIFF
--- a/xray/segment.go
+++ b/xray/segment.go
@@ -286,13 +286,13 @@ func (seg *Segment) Close(err error) {
 		seg.addError(err)
 	}
 
-	seg.Unlock()
-
 	// If segment is dummy we return
 	if seg.Dummy {
+		seg.Unlock()
 		return
 	}
 
+	seg.Unlock()
 	seg.send()
 }
 
--- a/xray/segment.go
+++ b/xray/segment.go
@@ -286,12 +286,13 @@ func (seg *Segment) Close(err error) {
 		seg.addError(err)
 	}
 
+	seg.Unlock()
+
 	// If segment is dummy we return
 	if seg.Dummy {
 		return
 	}
 
-	seg.Unlock()
 	seg.send()
 }
 


xray/segment.go
```diff
@@ -286,12 +286,13 @@ func (seg *Segment) Close(err error) {
 		seg.addError(err)
 	}
 
+	seg.Unlock()
+
 	// If segment is dummy we return
 	if seg.Dummy {
 		return
```
bhautikpip, line 293: I would suggest adding seg.Unlock() before this return statement because adding seg.Unlock before this block could cause data race while accessing the seg.Dummy field. Also, would you mind adding the same change in the CloseAndStream method.  (https://github.com/aws/aws-xray-sdk-go/blob/master/xray/segment.go#L319)

bhautikpip: So basically there will be 2 seg.Unlock call.

skaggmannen: `CloseAndStream` uses a defer for Unlock so it should be fine. From what I can tell it's not possible to do the same in `Close()` because of the call to `send()`, right?

bhautikpip: Yeah. Right! Thanks.

JesseObrien:
@bhautikpip is there a timeline for this fix to be released at all via point release? We've had this creep into our stack in the last two days causing production outages.

awssandra:
Hi JesseObrien,

We're expediting this fix, we'll be posting a new version very shortly.

JesseObrien:
@awssandra very much appreciated!! Thanks!

bhautikpip:
Hi ,

I recently released v1.0.1 which includes this fix. That should resolve most issues. Thanks for the patience!

PR #235, opened 2020-06-04
Added IMDS_V2 support

bhautikpip:
*Issue #, if available:*

*Description of changes:*

Added IMDS_V2 support.

### DIFF
--- a/awsplugins/ec2/ec2.go
+++ b/awsplugins/ec2/ec2.go
@@ -9,16 +9,35 @@
 package ec2
 
 import (
-	"github.com/aws/aws-sdk-go/aws/ec2metadata"
-	"github.com/aws/aws-sdk-go/aws/session"
+	"bytes"
+	"encoding/json"
 	"github.com/aws/aws-xray-sdk-go/internal/logger"
 	"github.com/aws/aws-xray-sdk-go/internal/plugins"
+	"net/http"
 )
 
 // Origin is the type of AWS resource that runs your application.
 const Origin = "AWS::EC2::Instance"
 
-// Init activates EC2Plugin at runtime.
+type Metdata struct {
+	AccountId string
+	Architecture string
+	AvailabilityZone string
+	BillingProducts string
+	DevpayProductCodes string
+	MarketplaceProductCodes string
+	ImageId string
+	InstanceId string
+	InstanceType string
+	KernelId string
+	PendingTime string
+	PrivateIp string
+	RamdiskId string
+	Region string
+	Version string
+}
+
+//Init activates EC2Plugin at runtime.
 func Init() {
 	if plugins.InstancePluginMetadata != nil && plugins.InstancePluginMetadata.EC2Metadata == nil {
 		addPluginMetadata(plugins.InstancePluginMetadata)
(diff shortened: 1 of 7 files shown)


awsplugins/ec2/ec2.go
```diff
@@ -9,35 +9,115 @@
...
+func getToken(imdsURL string, client *http.Client) (string, error) {
+	ttlHeader := "X-aws-ec2-metadata-token-ttl-seconds"
+	defaultTTL := "60"
+	tokenURL := imdsURL + "api/token"
+
+	req, _ := http.NewRequest("PUT", tokenURL, nil)
+	req.Header.Add(ttlHeader, defaultTTL)
+	resp, err := client.Do(req)
+	if err != nil {
+		return "fallback", err
```
willarmiros, line 91: Do you think we could return `nil` instead of a hardcoded string when the token request fails? It seems unnecessarily complicated to me to have a specific string for an invalid token.

awsplugins/ec2/ec2.go
```diff
@@ -9,35 +9,115 @@
...
+	if token != "fallback" {
+		req.Header.Add(metadataHeader, token)
+	}
+
+	resp, err := client.Do(req)
+	if err != nil {
+		return resp, err
+	}
+
+	return resp, err
```
willarmiros, line 122: can you just replace this whole block with:
```go
return client.Do(req)
```

awsplugins/ec2/ec2.go
```diff
@@ -9,35 +9,115 @@
...
+func getMetadata(imdsURL string, client *http.Client, token string) (*http.Response, error) {
+	var metadataHeader string
+	metadataURL := imdsURL + "dynamic/instance-identity/document"
+
+	if token != "fallback" {
+		metadataHeader = "X-aws-ec2-metadata-token"
+	}
+
+	req, _ := http.NewRequest("GET", metadataURL, nil)
+	if token != "fallback" {
```
willarmiros, line 113: Can we combine this with the above `if token != "fallback"` for conciseness?

awsplugins/ec2/ec2.go
```diff
@@ -9,35 +9,115 @@
...
+	resp, err := getMetadata(imdsURL, client, token)
 	if err != nil {
 		logger.Errorf("Unable to read EC2 instance metadata: %v", err)
 		return
 	}
 
-	pluginmd.EC2Metadata = &plugins.EC2Metadata{InstanceID: doc.InstanceID, AvailabilityZone: doc.AvailabilityZone}
+	buf := new(bytes.Buffer)
+	if _, err := buf.ReadFrom(resp.Body); err != nil {
+		logger.Errorf("Error while reading data from response buffer: %v", err)
```
willarmiros, line 69: return after error

awsplugins/ec2/ec2.go
```diff
@@ -9,35 +9,115 @@
...
+	}
+	metadata := buf.String()
+
+	if err := json.Unmarshal([]byte(metadata), &instanceData); err != nil {
+		logger.Errorf("Error while unmarshal operation: %v", err)
```
willarmiros, line 74: return after error

awsplugins/ec2/ec2.go
```diff
@@ -9,35 +9,115 @@
...
+	}
+
+	pluginmd.EC2Metadata = &plugins.EC2Metadata{InstanceID: instanceData.InstanceID, AvailabilityZone: instanceData.AvailabilityZone}
```
willarmiros, line 77: Can you include the `instance_size` and `ami_id` attributes as well? See: https://github.com/aws/aws-xray-sdk-node/pull/300

awsplugins/ec2/ec2.go
```diff
@@ -9,35 +9,115 @@
...
+
 	"github.com/aws/aws-xray-sdk-go/internal/logger"
 	"github.com/aws/aws-xray-sdk-go/internal/plugins"
 )
 
 // Origin is the type of AWS resource that runs your application.
 const Origin = "AWS::EC2::Instance"
 
-// Init activates EC2Plugin at runtime.
+type Metdata struct {
```
willarmiros, line 23: For performance reasons, we probably shouldn't record the attributes we don't use. We should only need the 4 of these we record.

awsplugins/ec2/ec2.go
```diff
@@ -9,35 +9,115 @@
...
 func addPluginMetadata(pluginmd *plugins.PluginMetadata) {
-	session, e := session.NewSession()
-	if e != nil {
-		logger.Errorf("Unable to create a new ec2 session: %v", e)
-		return
+	var instanceData Metdata
+	imdsURL := "http://169.254.169.254/latest/"
+
+	client := &http.Client{
+		Transport: http.DefaultTransport,
```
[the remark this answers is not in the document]
bhautikpip: Yes, if we instrument http calls before actually calling the Init() method.

awsplugins/ec2/ec2.go
bhautikpip: Yes. Makes sense!

awsplugins/ec2/ec2.go
```diff
@@ -9,35 +9,115 @@
...
+	}
+
+	token, err := getToken(imdsURL, client)
+	if err != nil {
+		logger.Errorf("Unable to fetch EC2 instance metadata token fallback to IMDS V1: %v", err)
```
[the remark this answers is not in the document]
bhautikpip: right.

awsplugins/ec2/ec2.go
bhautikpip: DOne.

awsplugins/ec2/ec2.go
bhautikpip: Done.

awsplugins/ec2/ec2.go
bhautikpip: Right.

awsplugins/ec2/ec2.go
bhautikpip: Unfortunately we have to return a string. I was thinking returning an empty string but then thought this would be better than empty string as we can identify case when there's a token failure and also can be useful in unit testing.

awsplugins/ec2/ec2.go
```diff
@@ -9,35 +9,115 @@
...
+	req, _ := http.NewRequest("PUT", tokenURL, nil)
+	req.Header.Add(ttlHeader, defaultTTL)
+	resp, err := client.Do(req)
+	if err != nil {
+		return "fallback", err
+	}
+
+	buf := new(bytes.Buffer)
+	if _, err := buf.ReadFrom(resp.Body); err != nil {
+		logger.Errorf("Error while reading data from response buffer: %v", err)
```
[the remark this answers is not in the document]
bhautikpip: yes but this would give false impression because this is not token retrieve failure. this is actually parsing failure. we don't want to fallback to IMDSv1 in the case of unable to parse token data right ?

awsplugins/ec2/ec2.go
bhautikpip: Done.

awsplugins/ec2/ec2.go
bhautikpip: Done.

awsplugins/ec2/ec2.go
willarmiros: Ok fair enough.

awsplugins/ec2/ec2.go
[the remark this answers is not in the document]
willarmiros: Well if we don't return here what will happen? It will return a corrupted string token that the plugin will attempt to use for a V2 request, which could cause unexpected behavior. 

I think returning the `"fallback", err` is acceptable here since we already have an invalid token so we know V2 won't work. There's no harm in just ignoring the token and trying V1 IMO, worst case we just log another failure if V1 doesn't work either.

awsplugins/ec2/ec2.go
anuraaga: I personally find this hardcoded string to be weird too. Either we got a token or we didn;'t, modeling the latter with empty string seems equivalent even for unit tests without introducing this magic string, right?

anuraaga: If you want to stick with the string at least extract a constant - but recommend empty string.

awsplugins/ec2/ec2.go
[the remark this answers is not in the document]
anuraaga: Agree that if the token can't be read (seems like it'd be for a connection reset when reading a very short string, so very unlikely but yeah can happen), 

- Optionally retry token fetch with upper limit / backoff
- Send request without token and it will fail if V1 is disabled but succeed otherwise

Sending a bad token should cause the metadata fetch to be guaranteed to fail, so there's no point to do it, we should either fallback, retry, or give up in our code.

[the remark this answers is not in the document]
bhautikpip: Okay. Makse sense! I will return an empty string here too.  so that it fallbacks to v1.

awsplugins/ec2/ec2_test.go
```diff
@@ -0,0 +1,122 @@
...
+	"io/ioutil"
+	"net/http"
+	"net/http/httptest"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+const metadata = `{
+  "availabilityZone" : "us-west-2a",
```
anuraaga, line 13: I think the previous version with a more representative doc was better, it does add a bit more test coverage by making sure unexpected keys don't break anything

PR #239, opened 2020-06-25
Fix typo (Metdata -> Metadata)

aereal:
*Issue #, if available:*

*Description of changes:*

`ec2.Metdata` type that introduced by #235 is maybe typo of `ec2.Metadata`.

### DIFF
--- a/awsplugins/ec2/ec2.go
+++ b/awsplugins/ec2/ec2.go
@@ -105,3 +105,12 @@ func getMetadata(imdsURL string, client *http.Client, token string) (*http.Respo
 
 	return client.Do(req)
 }
+
+// Metdata is deprecated.
+// It will be removed in the future but kept only for backward compatibility.
+type Metdata struct {
+	AvailabilityZone string
+	ImageID          string
+	InstanceID       string
+	InstanceType     string
+}
--- a/awsplugins/ec2/ec2.go
+++ b/awsplugins/ec2/ec2.go
@@ -106,8 +106,9 @@ func getMetadata(imdsURL string, client *http.Client, token string) (*http.Respo
 	return client.Do(req)
 }
 
-// Metdata is deprecated.
-// It will be removed in the future but kept only for backward compatibility.
+// Metdata represents IMDS response.
+//
+// Deprecated: Metdata exists for only backward compatibility.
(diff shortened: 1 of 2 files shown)


anuraaga:
Actually, on second type, let's leave `Metdata` as is, move it to the bottom of the file, and indicate it's deprecated, and add the new properly named, private struct. Since we don't return the type from an API, I think we can fix this without breaking semantic versioning

aereal:
@anuraaga 

Yeah, I have same concern about semantic versioning.

So changes I should make will be:

- add a new `metadata` private type that has same structure as `Metdata`
- indicates `Metdata` as deprecated

awsplugins/ec2/ec2.go
```diff
@@ -105,3 +105,12 @@ func getMetadata(imdsURL string, client *http.Client, token string) (*http.Respo
 
 	return client.Do(req)
 }
+
+// Metdata is deprecated.
+// It will be removed in the future but kept only for backward compatibility.
```
anuraaga, line 110: Thanks! Can you go ahead and follow the convention here?

https://github.com/golang/go/wiki/Deprecated

aereal: Oops, fixed by daa47e1

awsplugins/ec2/ec2_test.go
```diff
@@ -9,7 +9,7 @@ import (
 	"github.com/stretchr/testify/assert"
 )
 
-const metadata = `{
+const testMetadata = `{
```
anuraaga, line 12: LGTM thanks!

aereal:
@anuraaga

Thanks for your quick review.

Fixed deprecated comment convention and test failing that occurred by name conflicts.

awsplugins/ec2/ec2.go
```diff
@@ -105,3 +105,13 @@ func getMetadata(imdsURL string, client *http.Client, token string) (*http.Respo
 
 	return client.Do(req)
 }
+
+// Metdata represents IMDS response.
+//
+// Deprecated: Metdata exists for only backward compatibility.
```
anuraaga, line 111: `only for`

Issue #242, opened 2020-07-20
xray.BeginSegment does not honor sampling rule


kevioke:
Hi everyone. I'm currently trying to debug an issue where it seems like xray is sampling more requests than it should. I'm currently suspecting that my use of `xray.BeginSegment` is not properly honoring the custom centralized sampling rule that I've set up.

# Setup
package versions:
`go version go1.13.9 darwin/amd64`
` github.com/aws/aws-xray-sdk-go v1.0.1`

Custom centralized sampling rule with priority 1 named **test-sampling-rule** (I set this up in the AWS console):
```
Service name equals some-other-service-name
Service type matches *
Host matches *
Resource ARN matches *
HTTP method matches *
URL path matches *
```
Default sampling rule
```
Service name matches *
Service type matches *
Host matches *
Resource ARN matches *
HTTP method matches *
URL path matches *
```

My reproduction code simply produces a Segment every 5 seconds:
```go
package main

import (
	"context"
	"fmt"
	"os"
	"time"

	"github.com/aws/aws-xray-sdk-go/xray"
	"github.com/aws/aws-xray-sdk-go/xraylog"
)

func main() {
	xray.SetLogger(xraylog.NewDefaultLogger(os.Stderr, xraylog.LogLevelDebug))

	for i := 0; i < 100; i++ {
		ctx := context.Background()
		ctx, seg := xray.BeginSegment(ctx, "my-test-service")
		seg.Close(nil)
		time.Sleep(5 * time.Second)
		fmt.Printf("Sent %d segment(s)\n", i)
	}

	fmt.Println("done")
}

```

# Partial Output
```
2020-07-20T08:50:21-07:00 [DEBUG] Beginning segment named my-test-service
2020-07-20T08:50:21-07:00 [DEBUG] Determining ShouldTrace decision for:
	host:
	path:
	method:
	servicename:
	servicetype:
2020-07-20T08:50:21-07:00 [DEBUG] Applicable rule: test-sampling-rule
2020-07-20T08:50:21-07:00 [DEBUG] Sampling target has expired for rule test-sampling-rule. Borrowing a request.
2020-07-20T08:50:21-07:00 [DEBUG] SamplingStrategy decided: true
2020-07-20T08:50:21-07:00 [DEBUG] Closing segment named my-test-service
[...]

PR #244, opened 2020-07-24
Fix xray.BeginSegment to honor sampling rule

kevioke:
Fixes https://github.com/aws/aws-xray-sdk-go/issues/242

*Description of changes:*
Requests that have empty fields should not apply to sampling rules that require specific patterns to match. This change removes the `OR` condition on having empty fields short circuit any sampling rule pattern match.

In addition, this change also passes the `ServiceName` to the ShouldTrace in the BeginSegmentWithSampling function so that sampling rules can be applied more accurately with traces that only specify ServiceNames.


strategy/sampling/sampling_rule_test.go
```diff
@@ -343,3 +343,145 @@ func BenchmarkCentralizedRule_Sample(b *testing.B) {
...
+	type testCase struct {
+		testName string
+		rule     *CentralizedRule
+		request  *Request
+		expected bool
+	}
+
+	testCases := []testCase{
+		{
+			testName: "if the request is empty and the rule has a specific ServiceName, AppliesTo should return false",
```
bhautikpip, line 357: Can we add one test case here, where all the values of rule are populated like URLPath, ServiceName, ServiceType, Host and httpMethod. Now we pass the empty request, AppliesTo should return false ?

xray/segment.go
```diff
@@ -83,8 +83,8 @@ func BeginSegmentWithSampling(ctx context.Context, name string, r *http.Request,
 	}
 
 	if r == nil || traceHeader == nil {
-		// Sampling strategy fallbacks to default sampling in the case of directly using BeginSegment API without any instrumentation
-		sd := seg.ParentSegment.GetConfiguration().SamplingStrategy.ShouldTrace(&sampling.Request{})
+		// No header or request information provided so we can only evaluate sampling based on the serviceName
```
bhautikpip, line 86: Also, can we include this as a Note in README file so that sampling behavior is very clear when customer use BeginSegment API directly. (since we are evaluating only based on service name here)

strategy/sampling/sampling_rule.go
```diff
@@ -31,19 +31,19 @@ type Properties struct {
...
-		(path == "" || pattern.WildcardMatchCaseInsensitive(p.URLPath, path)) &&
-		(method == "" || pattern.WildcardMatchCaseInsensitive(p.HTTPMethod, method))
+	return pattern.WildcardMatchCaseInsensitive(p.Host, host) &&
+		pattern.WildcardMatchCaseInsensitive(p.URLPath, path) &&
+		pattern.WildcardMatchCaseInsensitive(p.HTTPMethod, method)
 }
 
 // AppliesTo returns true if the sampling rule matches against given sampling request. False Otherwise.
 // Assumes lock is already held, if required.
 func (r *CentralizedRule) AppliesTo(request *Request) bool {
```
willarmiros, line 41: Changing this to exclude absent request attributes would lead to behavior inconsistent with other SDKs, for example our [node SDK](https://github.com/aws/aws-xray-sdk-node/blob/master/packages/core/lib/middleware/sampling/sampling_rule.js#L38-L42) and [Python SDK](https://github.com/aws/aws-xray-sdk-python/blob/master/aws_xray_sdk/core/sampling/sampling_rule.py#L45-L49). I understand the ask to change this, but changing the behavior like this would be a fairly consequential, arguably breaking, change.

@kevioke I believe the change you made to the `beginSegment` API should be sufficient to resolve #242 since the `serviceName` will no longer be absent. Our team will have to discuss this change which may delay this PR quite a bit. If you'd like to remove this change and only include the `beginSegment` change that could expedite this getting released.

kevioke: @willarmiros thanks for the reply! I had a hunch this empty request logic was present for a reason, just didn't know quite why. In any case, I like your proposal of just including the changes to `beginSegment` to speed this PR along, and I believe you're correct that it should work for my use cases since now I can specify a rule with a ServiceName.

I'll update the PR later today or tomorrow.

README.md
```diff
@@ -137,6 +137,7 @@ func init() {
 ```
 
 **Start a custom segment/subsegment**
+Note that the `xray.BeginSegment` function currently uses only the service name parameter to determine the applicability of any sampling rules.
```
bhautikpip, line 140: Hi can we rephrase this to something like customers using xray.BeginSegment API directly will only be able to evaluate sampling rules based on service name ? It's important to show that this behavior is not applicable for instrumented requests.

kevioke: That is much clearer :)

I'll update it accordingly

Issue #265, opened 2020-12-06
Can't run Github Actions due to deprecated set-env syntax


sawadashota:
I created PR #264 but all Github Actions jobs went fail because this project uses deprecated set-env syntax.
https://github.blog/changelog/2020-10-01-github-actions-deprecating-set-env-and-add-path-commands/

bhautikpip:
Sure. Thanks for creating an issue. I will submit a PR to address test case failures.

PR #267, opened 2020-12-07
ci: Fix set-env and add-path syntax

sawadashota:
*Issue #265*

*Description of changes:*

Fix `set-env` and `add-path` because these are deprecated and has looked dropped.


bhautikpip (approved):
LGTM

PR #268, opened 2020-12-08
Add test for respecting manually set aws sdk headers

willarmiros:
*Description of changes:*
Adds a unit test to verify that we respect if a customer manually sets their Trace Header for an AWS SDK request, and don't overwrite it with our own header. This is for cases like SFN which are long-running and may require the customer to pass in custom headers which correspond to a particular in-flight execution.


willarmiros:
CI failing because of deprecated commands, will rerun after fix.

willarmiros:
Discussing with AWS SDK team whether this PR reflects desired behavior - will need to wait on merging.

Issue #231, opened 2020-05-25
Support custom database label for sql.db tracing


cgstag:
Right now, when using `xray.SQLContext(driver,DSN)`, in order to open a DB connection with tracing, the DSN is parsed and the dbname of the DSN is used to inject into the Capture function, used under the hood to trace the calls.
Exemple in the QueryContext method :
```golang
err = Capture(ctx, stmt.attr.dbname, func(ctx context.Context) error {
			stmt.populate(ctx)
			var err error
			result, err = stmt.Stmt.Query(dargs)
			return err
		})
```

However, in the case where two Databases hold the same dbname, but are different databases with different DSN (different host, different tables...) XRay will think it is the same database in the service map.

It would be really helpful to be able to set a custom label (that can default to attr.dbname).
I can contribute to make that feature happen, but I really dont know what is the best way to achieve this.

bhautikpip:
Hi @cgstag ,

Thanks for opening this issue. We are looking into this issue to see what could be the best way to address this issue. Meanwhile if you have any ideas please post here so that we can also discuss that. Thank you for the patience!

cgstag:
Hello @bhautikpip , has this topic been discussed internally ?
I think the quickest / easiest way would be instead of using the `dbname` to just use the entire `dsn` or a combination of `host` + `dbname`. Another way is to provide another constructor with Options, and providing the display label as one of the options.

I tried to play around DSN parameters but not only is it dependant on the drivers used, it also can mess up the parsing altogether, so I think its a no go.

bhautikpip:
Hi @cgstag ,

I agree with you. `dbname@host `looks simpler and makes sense to me. I looked at the implementation of [X-Ray Java SDK ](https://github.com/aws/aws-xray-sdk-java/blob/c8ad3c1454e5ff9c1da875b8fe15b5a855859dec/aws-xray-recorder-sdk-sql-mysql/src/main/java/com/amazonaws/xray/sql/mysql/TracingInterceptor.java#L129) and 
[...]

PR #273, opened 2021-01-07
Dbname issue

bhautikpip:
*Issue #, if available:*

*Description of changes:*

Appended host to the subsegment name in order to give different identity of 2 database with same name.

1. This case will only work for known DSL where host is parsable

So for know DSL subsegment will look like: `dbname@host` as per issue https://github.com/aws/aws-xray-sdk-go/issues/231


srprash (approved):
Looks good.

PR #274, opened 2021-01-14
Release workflow

bhautikpip:
*Issue #, if available:*

*Description of changes:*

Automated release process for X-Ray Go SDK

Tested this workflow on my fork: https://github.com/bhautikpip/aws-xray-sdk-go/runs/1705266557?check_suite_focus=true

The workflow released v1.3.0 on my fork here - https://github.com/bhautikpip/aws-xray-sdk-go/releases/tag/v1.3.0


anuraaga (approved):
I'm wondering how useful this is vs just creating releases in the UI. I guess there's a bit less chance of error though so maybe it's fine.

bhautikpip:
> I'm wondering how useful this is vs just creating releases in the UI. I guess there's a bit less chance of error though so maybe it's fine.

Yes the major motivation behind is we still have to clone repo and push tags and for that we need to perform release processes so we can reduce operational burden by having an automated workflow for this simple case as well :)

.github/workflows/release.yml
```diff
@@ -0,0 +1,31 @@
...
+        required: true
+
+jobs:
+  release:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout master branch
+        uses: actions/checkout@v2
+
+      - name: Initialize Github name and email
```
anuraaga, line 16: Ah - I dont think this is needed to create a tag (it would be to create a commit)

bhautikpip: Good catch! Yeah I will remove it.

PR #283, opened 2021-03-19
Add smoke test of distribution channel.

anuraaga:
*Issue #, if available:*

*Description of changes:*
Just simple test for now without metric publishing.


.github/workflows/continuous-monitoring.yml
```diff
@@ -0,0 +1,19 @@
+name: Continuous monitoring of distribution channels
+on:
+  workflow_dispatch:
```
bhautikpip, line 3: Do we need `workflow_dispatch` here? I think this should be fine as well.

```
name: Do things every 6 hours
on:
  schedule:
    - cron: "* */6 * * *"
```

bhautikpip (approved):
Added one minor comment otherwise looks good :) Thanks for adding this!

.github/workflows/continuous-monitoring.yml
anuraaga: I think it's useful to be about to manually run any cron job from ui or you always have to wait for the timer.

PR #285, opened 2021-03-22
Publish metric on distribution availability

bhautikpip:
*Issue #, if available:*

*Description of changes:*

Publish metric on distribution availability

### DIFF
--- a/.github/workflows/continuous-monitoring.yml
+++ b/.github/workflows/continuous-monitoring.yml
@@ -2,7 +2,7 @@ name: Continuous monitoring of distribution channels
 on:
   workflow_dispatch:
   schedule:
-    - cron:  '0 */6 * * *'
+    - cron:  '0 */1 * * *'
 
 jobs:
   smoke-tests:
--- a/.github/workflows/continuous-monitoring.yml
+++ b/.github/workflows/continuous-monitoring.yml
@@ -12,6 +12,13 @@ jobs:
       - name: Checkout Repository
         uses: actions/checkout@v2
       
+      - name: Configure AWS Credentials
+        uses: aws-actions/configure-aws-credentials@v1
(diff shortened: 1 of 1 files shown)


.github/workflows/continuous-monitoring.yml
```diff
@@ -11,9 +11,22 @@ jobs:
...
       - name: Run distribution test
+        id: distribution-availability
         run: go test distributioncheck
         working-directory: integration-tests/distributioncheck
+      
+      - name: Publish metric on X-Ray Go SDK distribution availability
+        if: ${{ always() }}
+        run: |
+          if [[ "${{ steps.distribution-availability.outcome }}" == "failure" ]]; then
+            aws cloudwatch put-metric-data --metric-name XRayGoSDKDistributionAvailability --dimensions failure=rate --namespace MonitorSDK --value 1 --timestamp $(date +%s)
```
willarmiros, line 28: How do we know the AWS CLI is installed?

willarmiros (approved):
Just 1 question

.github/workflows/continuous-monitoring.yml
bhautikpip: Github instance installs the latest version of AWS CLI by default.

anuraaga: Don't we need to add aws login? Also what account are we going to use? Ideally we use https://github.com/aws-observability/aws-otel-ops to set up the secrets.

bhautikpip: we need to configure aws credentials on Github instance. We are using the testing account for this so we can use secrets that are set up for integration test workflow no need to set up another secrets.

anuraaga (approved):

PR #292, opened 2021-04-01
Support instrumenting unary grpc

tony84727:
*Issue #, if available:*
#110 

*Description of changes:*
Continue @sawadashota's effort (#264) to implement grpc support.


README.md
```diff
@@ -276,6 +276,30 @@ func HandleRequest(ctx context.Context, name string) (string, error) {
...
+        xray.UnaryClientInterceptor(serverAddr),
+    )),
+)
+```
+
+**grpc server**
+
+```go
+grpcServer := grpc.NewServer(
+    grpcmiddleware.WithUnaryServerChain(
```
bhautikpip, line 297: We should probably include below imports here to give clear idea about `grpcmiddleware` package

```
grpc_middleware "github.com/grpc-ecosystem/go-grpc-middleware"
```

xray/grpc.go
```diff
@@ -0,0 +1,143 @@
+package xray
+
+import (
+	"context"
+	"errors"
+	"github.com/golang/protobuf/proto"
```
bhautikpip, line 6: Can we format this file to avoid review dog warnings?

README.md
```diff
@@ -276,6 +276,30 @@ func HandleRequest(ctx context.Context, name string) (string, error) {
 }
 ```
 
+`aws-xray-sdk-go` doesn't currently support streaming grpc call.
+
+**grpc client**
+
+```go
+conn, err := grpc.Dial(
```
bhautikpip, line 284: Can we add a little description here on how below grpc client and server instrumentation works ?

README.md
tony84727: `grpc` had added similar chaining methods so we don't need to use `grpc_middleware` now.

To apply the interceptors, we can use [`grpc.UnaryInterceptor`](https://pkg.go.dev/google.golang.org/grpc#UnaryInterceptor) r [`grpc.WithUnaryInterceptor`](https://pkg.go.dev/google.golang.org/grpc#WithUnaryInterceptor) directly. So I simplify the examples a bit and add comments to show how to apply multiple interceptors.

xray/grpc.go
tony84727: Ran the `go fmt` for the file, those warning should be fixed.

README.md
tony84727: Added a short description to point out the instrumentation is done by applying `xray.UnaryServerInterceptor` or `xray.UnaryClientInterceptor`.

xray/grpc.go
```diff
@@ -0,0 +1,143 @@
...
+				ct.subsegments.GotConn(nil, err)
+			}
+
+			return err
+		})
+	}
+}
+
+// UnaryServerInterceptor provides gRPC unary server interceptor.
+func UnaryServerInterceptor(ctx context.Context, sn SegmentNamer) grpc.UnaryServerInterceptor {
```
tony84727, line 55: AFAIK, the `ctx` here is only used to pass in the value of  `RecorderContextKey`, which should be a `Config`.

Normally, it's a `nil` so the recorder uses the global config.

I can see that for HTTP instrumentation, there's two types of wrapper:
1. [`xray.Handler`](https://pkg.go.dev/github.com/aws/aws-xray-sdk-go/xray#Handler)
2. [`xray.HandlerWithContext`](https://pkg.go.dev/github.com/aws/aws-xray-sdk-go/xray#HandlerWithContext)

Maybe we should break the interceptor into two signatures as well? One without context and another one with context.

bhautikpip: I think that's a fair call out. We can have two types of wrapper here as well with `ctx` and without `ctx`

tony84727: added `UnaryServerInterceptorWithContext`.

xray/grpc.go
```diff
@@ -0,0 +1,152 @@
...
+		seg.GetHTTP().GetRequest().ClientIP, seg.GetHTTP().GetRequest().XForwardedFor = clientIPFromGrpcMetadata(md)
+		seg.GetHTTP().GetRequest().URL = requestURL.String()
+		seg.GetHTTP().GetRequest().Method = info.FullMethod
+		if len(md.Get("user-agent")) == 1 {
+			seg.GetHTTP().GetRequest().UserAgent = md.Get("user-agent")[0]
+		}
+		seg.Unlock()
+
+		resp, err = handler(ctx, req)
+		recordContentLength(seg, resp)
```
bhautikpip, line 96: I think we also probably should include setting up response header and set `_x_amzn_trace_id` value. Reference can be taken from here - https://github.com/tony84727/aws-xray-sdk-go/blob/support_grpc/xray/handler.go#L133. If we are not generating response header and add it as a value to the `_x_amzn_trace_id` key then downstream services will keep on sampling even if we did sampling on upstream service.

tony84727: Calling `grpc.SendHeader` to send the response trace header. `grpc.SendHeader` might returns an error, I'm using [`multierr`](https://github.com/uber-go/multierr) to wrap multiple errors together. Not sure whether it's a good idea. Or should we just ignore the error and fail silently?

xray/grpc.go
```diff
@@ -0,0 +1,174 @@
...
+	"github.com/aws/aws-xray-sdk-go/header"
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/metadata"
+)
+
+// UnaryClientInterceptor provides gRPC unary client interceptor.
+func UnaryClientInterceptor(host string) grpc.UnaryClientInterceptor {
+	return func(ctx context.Context, method string, req, reply interface{}, cc *grpc.ClientConn, invoker grpc.UnaryInvoker, opts ...grpc.CallOption) error {
+		return Capture(ctx, host, func(ctx context.Context) error {
+			seg := GetSegment(ctx)
```
anuraaga, line 27: This looks like it is getting the server segment and is then mutated below. The unit test seems sane though, could you point out where the subsegment is being created for the client request?

xray/grpc.go
```diff
@@ -0,0 +1,174 @@
...
+			if e != nil {
+				return e
+			}
+			md := metadata.Pairs(TraceIDHeaderKey, seg.DownstreamHeader().String())
+			ctx2 := metadata.NewOutgoingContext(httptrace.WithClientTrace(ctx, ct.httpTrace), md)
+
+			seg.Lock()
+			seg.Namespace = "remote"
+			seg.GetHTTP().GetRequest().URL = "grpc://" + host + method
+			seg.GetHTTP().GetRequest().Method = method
```
anuraaga, line 42: This needs to be an HTTP method, not gRPC method. It's always `POST`.

xray/grpc.go
```diff
@@ -0,0 +1,174 @@
...
+		}
+		name := sn.Name(host)
+
+		ctx = context.WithValue(ctx, RecorderContextKey{}, cfg)
+
+		var seg *Segment
+		ctx, seg = NewSegmentFromHeader(ctx, name, &http.Request{
+			Host:   host,
+			URL:    &requestURL,
+			Method: info.FullMethod,
```
anuraaga, line 86: Ditto this needs to be `POST`

xray/grpc.go
```diff
@@ -0,0 +1,174 @@
...
+		}, traceHeader)
+		defer seg.Close(nil)
+
+		seg.Lock()
+		seg.GetHTTP().GetRequest().ClientIP, seg.GetHTTP().GetRequest().XForwardedFor = clientIPFromGrpcMetadata(md)
+		seg.GetHTTP().GetRequest().URL = requestURL.String()
+		seg.GetHTTP().GetRequest().Method = info.FullMethod
```
anuraaga, line 93: `POST`

xray/grpc.go
```diff
@@ -0,0 +1,174 @@
...
+		}
+		seg.Unlock()
+
+		resp, err = handler(ctx, req)
+		if err != nil {
+			classifyErrorStatus(seg, err)
+		}
+		recordContentLength(seg, resp)
+		if headerErr := addResponseTraceHeader(ctx, seg, traceHeader); headerErr != nil {
+			err = multierr.Combine(err, headerErr)
```
anuraaga, line 105: Can we avoid adding this new library dependency on `multierr`?

README.md
```diff
@@ -280,6 +280,35 @@ func HandleRequest(ctx context.Context, name string) (string, error) {
...
+)
+```
+
+**gRPC Server**
+
+```go
+grpcServer := grpc.NewServer(
+    // use grpc.ChainUnaryInterceptor instead to apply multiple interceptors
+    grpc.UnaryInterceptor(
+        xray.UnaryServerInterceptor(context.TODO(), xray.NewFixedSegmentNamer("myApp")),
```
anuraaga, line 307:
```suggestion
        xray.UnaryServerInterceptor(xray.NewFixedSegmentNamer("myApp")),
```

xray/grpc.go
```diff
@@ -0,0 +1,174 @@
...
+
+// UnaryServerInterceptorWithContext provides gRPC unary server interceptor. xray.Config of given ctx will be inherited
+// by the instrumented context.
+func UnaryServerInterceptorWithContext(ctx context.Context, sn SegmentNamer) grpc.UnaryServerInterceptor {
+	cfg := GetRecorder(ctx)
+	return unaryServerInterceptorWithConfig(sn, cfg)
+}
+
+// UnaryServerInterceptor provides gRPC unary server interceptor.
+func UnaryServerInterceptor(sn SegmentNamer) grpc.UnaryServerInterceptor {
```
anuraaga, line 120: It would be nice to have a default interceptor that uses the gRPC service name as the segment name

xray/grpc.go
```diff
@@ -0,0 +1,174 @@
...
+	defer seg.Unlock()
+	grpcStatus, ok := status.FromError(err)
+	if !ok {
+		seg.Error = true
+		return
+	}
+	switch grpcStatus.Code() {
+	case codes.ResourceExhausted:
+		seg.Throttle = true
+	case codes.Internal, codes.Unimplemented, codes.DataLoss:
```
anuraaga, line 135: Error is like HTTP 4xx. This list is a good one to refer to for mapping

https://github.com/open-telemetry/opentelemetry-collector/pull/3110/files#diff-3c93ea7fd05dbbc4514087f190d5c07c0d125e606559d8cdd96923081058b63dL38

xray/grpc.go
tony84727: Hello, @anuraaga
I believe the subsegment is created by `Capture` at line 26, which call `BeginSubsegment`.

xray/grpc.go
tony84727: Agree. So I guess there will be four variants of the server interceptor?
1. `UnaryServerInterceptor()`
1. `UnaryServerInterceptorWithContext(ctx context.Context)`
1. `UnaryServerInterceptorWithNamer(sn SegmentNamer)`
1. `UnaryServerInterceptorWithContextAndNamer(ctx context.Context, sn SegmentNamer)`

anuraaga: Yeah - can you use the options pattern to have some symmetry with upstream gRPC and avoid the need for the explosion of factory methods?

https://github.com/grpc/grpc-go/blob/master/server.go#L118

xray/grpc.go
tony84727: > https://github.com/open-telemetry/opentelemetry-collector/pull/3110/files#diff-3c93ea7fd05dbbc4514087f190d5c07c0d125e606559d8cdd96923081058b63dL38

The mapping doesn't classify all the possible grpc status codes here:
https://github.com/grpc/grpc-go/blob/dc77d7ffe311f78f2e577572d984af3c0a8df82b/codes/codes.go

So, I'm referencing the mapping here:
https://github.com/googleapis/googleapis/blob/b29fb6b9fcdf3bc67487fa22ab7783915023041e/google/rpc/code.proto#L35

xray/grpc.go
tony84727: Fixed

xray/grpc.go
tony84727: Fixed

xray/grpc.go
tony84727: Fixed

xray/grpc.go
tony84727: Fixed

xray/grpc.go
tony84727: As described here https://github.com/aws/aws-xray-sdk-go/pull/292#discussion_r623120247
At this line there're two possible error sources:
1. The error returned by the service handler itself
1. The error returned by calling `grpc.SendHeader`

I'm not sure how to report multiple errors within an interceptor. Maybe change this to print a line of error message when failling to send headers? What do you think?

anuraaga: I think we can drop the error, it seems pretty rare. Alternatively a debug log message could be ok too. But yeah we shouldn't add it to the `err` here. It's not quite as important if `err` is actually set but imagine if its `nil` - then it means our interceptor changes the status of the actual gRPC request to an error which will probably cause problems in user code. The interceptor should only record issues but not introduce errors to the business logic like that.

tony84727: Okay, using the internal logger to print the debug message.

README.md
tony84727: Updated. Thanks.

README.md
```diff
@@ -280,6 +280,36 @@ func HandleRequest(ctx context.Context, name string) (string, error) {
...
+```
+
+**gRPC Server**
+
+```go
+grpcServer := grpc.NewServer(
+    // use grpc.ChainUnaryInterceptor instead to apply multiple interceptors
+    grpc.UnaryInterceptor(
+        xray.UnaryServerInterceptor(),
+        // or xray.UnaryServerInterceptor(xray.ServerInterceptorWithSegmentNamer(xray.NewFixedSegmentNamer("myApp"))) to use a custom segment namer
```
anuraaga, line 308: This looks a bit wrong now

xray/grpc.go
```diff
@@ -0,0 +1,208 @@
...
+	"github.com/golang/protobuf/proto"
+	"google.golang.org/grpc/status"
+
+	"github.com/aws/aws-xray-sdk-go/header"
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/metadata"
+)
+
+// UnaryClientInterceptor provides gRPC unary client interceptor.
+func UnaryClientInterceptor(host string) grpc.UnaryClientInterceptor {
```
anuraaga, line 25: One last point, how about we follow option pattern for client too and make host an option. By default, can use the service name here too

xray/grpc_test.go
```diff
@@ -282,6 +287,28 @@ func TestUnaryServerInterceptor(t *testing.T) {
...
+				),
+			),
+		)
+		client, closeFunc := newGrpcClient(context.Background(), t, lis)
+		defer closeFunc()
+		_, err := client.Ping(context.Background(), &pb.PingRequest{Value: "something", SleepTimeMs: 9999})
+		assert.NoError(t, err)
+		segment, err := td.Recv()
+		assert.NoError(t, err)
+		assert.Equal(t, "bufnet/mwitkow.testproto.TestService/Ping", segment.Name)
```
anuraaga, line 310: Oh sorry also this should only be the service name by default, definitely not method and we don't need host in the default (a gRPC service name is semantically similar to a segment name, like a host would be on HTTP). So it should just be `mwitkow.testproto.TestService`

tony84727: Okay. Fixed

xray/grpc.go
tony84727: Applied the options pattern to UnaryClientInterceptor. The original `host` here is used as the segment name and host of the recorded request URL. So I added two options:
1. Segment Nmaer
1. Host
Although I think it's a rare case to override the host?

README.md
tony84727: Do you mean there's a typo or I shouldn't use comments to mention the usage of options?

anuraaga: I think typo, it's `xray.UnaryServerINterceptor(xray.NewFixedSegmentNamer(...))` right?

xray/grpc.go
anuraaga: Ah - it looks like you found host in the interceptor parameters, `cc.Target`. If that's available than I agree there isn't much use case to override host. Having the segment namer option looks good, how about we just go ahead and remove the host option for now and add it if someone asks for it?

README.md
tony84727: By the options pattern and the current implementation, adding a segment namer to the unary server interceptor should look like this:
https://github.com/aws/aws-xray-sdk-go/blob/63d86848b24a520ab8fb049e5e5c0006bbcb7264/xray/grpc_test.go#L262-L265

xray/grpc.go
tony84727: 👍 Removed the host option.

xray/grpc.go
```diff
@@ -0,0 +1,256 @@
...
+		option.ctx = ctx
+	})
+}
+
+// ServerInterceptorWithSegmentNamer makes the interceptor use the segment namer to name the segment.
+func ServerInterceptorWithSegmentNamer(sn SegmentNamer) ServerInterceptorOption {
+	return newFuncServerInterceptorOption(func(option *serverInterceptorOption) {
+		option.segmentNamer = sn
+	})
+}
```
tony84727, line 228: Maybe we should rename this into `UnaryServerInterceptorWithSegmentNamer`? Or organize interceptors and options into sub-packages?

We might need to find a proper way to organize those options because interceptors of gRPC have four variants:
1. Unary / Client
2. Unary / Server

In the future, we might implement interceptors for gRPC streaming. Then we will have:
3. Stream / Client
4. Stream / Server

I can imagine, different types of interceptors will have different types of concerns and options.

README.md
anuraaga: Ah I don't think we need to be redundant with e.g. `ServerInterceptor`. Just `UnaryServerInterceptor(WithContext(ctx), WithSegmentNamer(...))` can be done I think, there's no difference between server / client versions of these options right? If we can simplify these names I think we're good to go

anuraaga (commented):
Sorry for the extreme delay in getting back, just a comment about the option names, I think we're pretty close.

anuraaga:
Also it could be great reference to look through the OpenTelemetry instrumentation for good patterns re: naming, etc.

https://github.com/open-telemetry/opentelemetry-go-contrib/blob/main/instrumentation/google.golang.org/grpc/otelgrpc/interceptor.go

They don't seem to have separate packages for unarystream / clientserver so I'm thinking it's relatively scalable.

tony84727:
> Ah I don't think we need to be redundant with e.g. ServerInterceptor. Just UnaryServerInterceptor(WithContext(ctx), WithSegmentNamer(...)) can be done I think, there's no difference between server / client versions of these options right? If we can simplify these names I think we're good to go

Merged Server/ClientInterceptorOption -> GrpcOption.

I also add a refactor that changes the API a little bit:
Changed the option `WithContext` -> `withRecorder` because the option was used to provide a custom `xray.Config` to the server handling context. We only propagate the `xray.Config` value of the given context. We're not propagating other values associated with the context. This should make the option's intent more clear. (Also, it seems to be a code smell to store a context inside a struct)

anuraaga (approved):
Thanks for the patience! @bhautikpip or @srprash can you also give it a look over and we'll merge?

bhautikpip (approved):
LGTM :)

PR #293, opened 2021-04-01
No op trace

bhautikpip:
*Issue #, if available:*

*Description of changes:*

This PR focuses on generating no-op trace id for un-sampled traces. The default behavior would be to generate secure id on sampled traces and no-op id on un-sampled traces. However, customer can configure to create secure id for all the traces using `AWS_XRAY_SECURE_RANDOM_ID` (to include trace id injection into logs use case). I have also added benchmark and test cases to test out this scenario.


xray/segment.go
```diff
@@ -45,10 +45,23 @@ func NewSegmentID() string {
 	return fmt.Sprintf("%02x", r)
 }
 
+func noOpTraceID() string {
+	return "0-00000000-000000000000000000000000"
```
anuraaga, line 49: noop still needs to start with `1-`

xray/segment.go
```diff
@@ -126,9 +139,28 @@ func BeginSegmentWithSampling(ctx context.Context, name string, r *http.Request,
...
 
+	// generates segment and trace id based on sampling decision and secure random env variable
+	idGeneration(seg)
+
 	return context.WithValue(ctx, ContextKey, seg), seg
 }
 
+func idGeneration(seg *Segment) {
+	secureRandom := os.Getenv("AWS_XRAY_SECURE_RANDOM_ID")
+	if secureRandom != "" && strings.ToLower(secureRandom) == "true" {
```
willarmiros, line 150: Can simplify this condition with adding an `|| seg.Sampled` here right?

xray/segment.go
```diff
@@ -126,9 +139,28 @@ func BeginSegmentWithSampling(ctx context.Context, name string, r *http.Request,
...
 	}
 
+	// generates segment and trace id based on sampling decision and secure random env variable
+	idGeneration(seg)
+
 	return context.WithValue(ctx, ContextKey, seg), seg
 }
 
+func idGeneration(seg *Segment) {
+	secureRandom := os.Getenv("AWS_XRAY_SECURE_RANDOM_ID")
```
willarmiros, line 149: Not sure how I feel about this env var name, especially since the secure random vs. fast random configuration, which is a different configuration than this, may be added in the future. How about `AWS_XRAY_UNSAMPLED_IDS`? Or `AWS_XRAY_NOOP_ID` (which is true by default, so the opposite conditions as we have currently)?

willarmiros (commented):
Would like to discuss env var name a bit more.

xray/segment.go
bhautikpip: Yeah, I like `AWS_XRAY_NOOP_ID` name and setting variable as false should generate secure random IDs even for un-sampled traces.

xray/segment.go
bhautikpip: Yes we can but not sure if that's a good idea. Ideally no-op id generation or not would be based on env variable value. If we include `|| seg.Sampled` here then I think we are using `seg.Sampled` to decide this behavior. Ideally if you look at the code it's a layered condition and I think this would be more safer.

xray/segment_test.go
```diff
@@ -246,3 +307,21 @@ func BenchmarkAddError(b *testing.B) {
 	seg.Sampled = false
 	seg.Close(nil)
 }
+
+func BenchmarkIdGeneration_secureTrue(b *testing.B) {
```
willarmiros, line 311: Can we change the name of these benchmarks?

willarmiros (approved):
LGTM, just a small comment.

willarmiros (approved):
additional changes lgtm

PR #295, opened 2021-04-26
feature: instrumentation for fasthttp

nicolascb:
Issue #133 

Instrumentation for [fasthttp](https://github.com/valyala/fasthttp).

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #296, opened 2021-04-27
Refactoring http trace

nicolascb:
I am refactoring this function to help future implementations, such as what was proposed in #295 .

This PR is following this recommendation: https://github.com/aws/aws-xray-sdk-go/pull/295#issuecomment-827692667


xray/segment.go
```diff
@@ -125,6 +127,17 @@ func BeginSegmentWithSampling(ctx context.Context, name string, r *http.Request,
 			logger.Debugf("SamplingStrategy decided: %t", seg.Sampled)
 			seg.AddRuleName(sd)
 		}
+
```
bhautikpip, line 130: Not sure if moving this piece of code here would be a good idea since `BeginSegmentWithSampling` method is pretty heavy and keeping this code in here would not make this API code more extensible since this piece of code would specifically be useful for http calls. My recommendation would be keep this code inside `htttpTrace` method only or we can create a helper function inside `handler.go` file.

xray/segment.go
```diff
@@ -638,3 +651,38 @@ func (seg *Segment) addError(err error) {
...
+// TraceHeaderID return trace id for http header
+func (seg *Segment) TraceHeaderID(traceHeader *header.Header) string {
+	seg.Lock()
+	defer seg.Unlock()
+
+	var respHeader bytes.Buffer
+	respHeader.WriteString("Root=")
+	respHeader.WriteString(seg.TraceID)
+
```
bhautikpip, line 663: Same comment for `TraceHeaderId` and `HTTPCapture` methods. Any specific reason for bringing this here? We can also add these methods inside `handler.go` file. Also, not sure if we want to expose this APIs, this can be better served as internal helper methods I think.

xray/segment_test.go
```diff
@@ -325,3 +326,156 @@ func BenchmarkIdGeneration_noOpFalse(b *testing.B) {
...
+	}
+	tests := []struct {
+		name    string
+		init    func(t *testing.T) *Segment
+		inspect func(r *Segment, t *testing.T) //inspects receiver after test run
+
+		args func(t *testing.T) args
+
+		want1 string
+	}{
```
bhautikpip, line 342: Also, we can move the test to `handler_test.go`. This is inline with above comments.

bhautikpip (commented):
Thanks for making this change. Added some comments :)

xray/segment.go
nicolascb: The reason for being in this file is because they are methods of the `Segment` structure, so I thought it would make sense to be in `segment.go`.

I'm changing this and I'm going to create helper functions in `handler.go`.

xray/handler.go
```diff
@@ -146,3 +148,54 @@ func btoi(b bool) int {
 	}
 	return 0
 }
+
+// traceHeaderID return trace id for http header
```
bhautikpip, line 152: Here we are actually generating value of key `_x_amzn_trace_id` which we are going to set in trace header. So, We can probably provide comment like `Generates value for _x_amzn_trace_id header key` or something similar to this. Also, can provide helper function name as `generateTraceIDHeaderValue` or something similar in this direction.

xray/handler.go
```diff
@@ -125,7 +127,7 @@ func httpTrace(seg *Segment, h http.Handler, w http.ResponseWriter, r *http.Requ
 
 	cLen, _ := strconv.Atoi(capturer.Header().Get("Content-Length"))
 	seg.GetHTTP().GetResponse().ContentLength = cLen
```
bhautikpip, line 129: Probably better if we can reduce this to one line like below. Also, I think we should add locks here.

```
seg.Lock()
seg.GetHTTP().GetResponse().ContentLength = strconv.Atoi(capturer.Header().Get("Content-Length"))
seg.Unlock()
```

bhautikpip (commented):
Some minor comments. Otherwise looks good :)

xray/handler.go
```diff
@@ -117,19 +117,43 @@ func Handler(sn SegmentNamer, h http.Handler) http.Handler {
 }
 
 func httpTrace(seg *Segment, h http.Handler, w http.ResponseWriter, r *http.Request, traceHeader *header.Header) {
-	seg.Lock()
+	httpCaptureRequest(seg, r)
+	traceID := traceHeaderID(seg, traceHeader)
```
bhautikpip, line 121: Here we can rename this variable name to  `traceIDHeaderValue`

nicolascb:
Thanks @bhautikpip!

I submitted the changes :)

xray/segment.go
```diff
@@ -125,6 +125,7 @@ func BeginSegmentWithSampling(ctx context.Context, name string, r *http.Request,
 			logger.Debugf("SamplingStrategy decided: %t", seg.Sampled)
 			seg.AddRuleName(sd)
 		}
+
```
bhautikpip, line 128: Nit: can we remove spacing from here?

xray/handler.go
```diff
@@ -117,49 +117,19 @@ func Handler(sn SegmentNamer, h http.Handler) http.Handler {
...
-	if capturer.status >= 400 && capturer.status < 500 {
-		seg.Error = true
-	}
-	if capturer.status == 429 {
-		seg.Throttle = true
-	}
-	if capturer.status >= 500 && capturer.status < 600 {
-		seg.Fault = true
-	}
+	seg.GetHTTP().GetResponse().ContentLength = cLen
```
bhautikpip, line 130: Can we remove `cLen` variable here and assign value directly?

bhautikpip (commented):
Sorry for the last min changes. posted some final nit picks.

nicolascb:
@bhautikpip review again, pls

bhautikpip (approved):
Looks good :)

PR #304, opened 2021-05-31
fix: replace segment name from env on BeginSegment

nicolascb:
It is not clear from the [official documentation](https://docs.aws.amazon.com/xray/latest/devguide/xray-sdk-go-configuration.html) that this variable is used only in HTTP situations:

```
Environment variables

You can use environment variables to configure the X-Ray SDK for Go. The SDK supports the following variables.

    AWS_XRAY_TRACING_NAME – Set the service name that the SDK uses for segments.
```

This PR applies the same logic so that the segment name is replaced by the environment variable `AWS_XRAY_TRACING_NAME`.

### DIFF
--- a/xray/handler.go
+++ b/xray/handler.go
@@ -13,7 +13,6 @@ import (
 	"context"
 	"net"
 	"net/http"
-	"os"
 	"strconv"
 	"strings"
 
@@ -36,9 +35,6 @@ type FixedSegmentNamer struct {
 // If the AWS_XRAY_TRACING_NAME environment variable is set,
 // its value will override the provided name argument.
 func NewFixedSegmentNamer(name string) *FixedSegmentNamer {
-	if fName := os.Getenv("AWS_XRAY_TRACING_NAME"); fName != "" {
-		name = fName
-	}
 	return &FixedSegmentNamer{
 		FixedName: name,
 	}
@@ -63,9 +59,6 @@ type DynamicSegmentNamer struct {
 
 // NewDynamicSegmentNamer creates a new dynamic segment namer.
 func NewDynamicSegmentNamer(fallback string, recognized string) *DynamicSegmentNamer {
-	if dName := os.Getenv("AWS_XRAY_TRACING_NAME"); dName != "" {
-		fallback = dName
-	}
 	return &DynamicSegmentNamer{
 		FallbackName:    fallback,
 		RecognizedHosts: recognized,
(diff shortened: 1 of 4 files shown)


nicolascb:
That makes sense? @bhautikpip

bhautikpip:
Yes, I think this check will be useful when we directly call `xray.BeginSegment()` API and for other middlewares like http and fasthttp handler we are checking this env var in [`xray.NewFixedSegmentNamer`](https://github.com/aws/aws-xray-sdk-go/blob/015dc2ef8d2202bebd329e97775375ba84075cd1/xray/handler.go#L38) method. So I think for parity with other checks this would be a good change to add.

bhautikpip:
Should we remove the check from here https://github.com/aws/aws-xray-sdk-go/blob/015dc2ef8d2202bebd329e97775375ba84075cd1/xray/handler.go#L39  and keep it only at `xray.BeginSegment()` method ? I think in the case of http handler if customer has used `AWS_XRAY_TRACING_NAME` then the same condition will be evaluated twice.

nicolascb:
> Should we remove the check from here
> 
> https://github.com/aws/aws-xray-sdk-go/blob/015dc2ef8d2202bebd329e97775375ba84075cd1/xray/handler.go#L39
> and keep it only at `xray.BeginSegment()` method ? I think in the case of http handler if customer has used `AWS_XRAY_TRACING_NAME` then the same condition will be evaluated twice.

Sorry @bhautikpip, but I did not understand.

If you remove this condition, what about situations where we don't use xray.BeginSegment?

As in this example: 
image

bhautikpip:
So basically `xray.Handler` method calls `xray.NewSegmentFromHeader` (https://github.com/aws/aws-xray-sdk-go/blob/master/xray/handler.go#L94) method  and that method calls `xray.BeginSegmentWithSampling` (https://github.com/aws/aws-xray-sdk-go/blob/9d2f59c14ad263ae30e16a342bdec1946f7bfc66/xray/segment.go#L77).  so basically middlewares like http end up calling `BeginSegmentWithSampling` method so if we keep this condition inside `BeginSegmentWithSampling` then we can remove this condition inside `NewFixedSegmentNamer`. Hope this makes sense!

nicolascb:
@bhautikpip Yes you are right.

I will work on it, thanks for the review

nicolascb:
Done https://github.com/aws/aws-xray-sdk-go/pull/304/commits/de8e0b9ea87e8e826ac3ed6e664608b303451635 @bhautikpip

bhautikpip (approved):
LGTM :)

PR #308, opened 2021-06-13
Instead of creating new context append to outgoing context

stijndehaes:
For example: when setting authorization headers to an
outgoing context these will be removed when using
new context. By appending these stay in the outgoing
context


anuraaga (approved):
Thanks!

PR #297, opened 2021-04-29
Adding AWS SDK v2 support

srprash:
*Issue #, if available:* #277

*Description of changes:*
**Work in progress:**

TODOs:
- [x] Middleware for intercepting requests and begin subsegment
- [ ] Add support for defualt and custom paramater allow-lists
- [ ] Unit tests
- [ ] Documentation

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #309, opened 2021-06-14
aws sdk v2 instrumentation support

bhautikpip:
*Issue #, if available:*

*Description of changes:*

Continuation of this PR: https://github.com/aws/aws-xray-sdk-go/pull/297


README.md
```diff
@@ -216,6 +216,47 @@ xray.AWS(dynamo.Client)
...
+	ctx, root := xray.BeginSegment(context.TODO(), "AWSSDKV2_Dynamodb")
+	defer root.Close(nil)
+
+	cfg, err := config.LoadDefaultConfig(ctx, config.WithRegion("us-west-2"))
+	if err != nil {
+		log.Fatalf("unable to load SDK config, %v", err)
+	}
+
+	// Instrumenting AWS SDK v2
+	xray.AppendMiddlewares(&cfg.APIOptions)
```
srprash, line 243: Can we change the method name from "ApendMiddlewares" to something that communicates AWS SDK v2 instrumentation clearly? For example "AddAWSV2Tracing" or something else.

xray/aws_v2.go
```diff
@@ -0,0 +1,92 @@
...
+		}
+
+		if resp.StatusCode >= 400 && resp.StatusCode <= 499 {
+			subseg.Error = true
+			if resp.StatusCode == 429 {
+				subseg.Throttle = true
+			}
+		} else if resp.StatusCode >= 500 && resp.StatusCode <= 599 {
+			subseg.Fault = true
+		}
```
srprash, line 83: This little code block can be its own private method. Makes it a bit cleaner.

srprash (approved):
Looks good. Just some minor suggestions.
Thanks!

xray/aws_v2.go
```diff
@@ -0,0 +1,83 @@
+// Copyright 2017-2021 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License"). You may not use this file except in compliance with the License. A copy of the License is located at
+//
+//     http://aws.amazon.com/apache2.0/
+//
+// or in the "license" file accompanying this file. This file is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the specific language governing permissions and limitations under the License.
+
+package xray
```
anuraaga, line 9: Can you move this into another package? We found the current pattern bloats binary size so we should aim for a package for each instrumentation

bhautikpip: I have added the separate module for this PR but this change would require a lot of re working because currently we release the repository as a single package which includes instrumentation packages and SDK. Now, If we want to create a separate module for each instrumentation package then I think that would require moving things around and basically separate out the SDK and instrumentation packages. Once we do that then, we can manage separate release of core SDK and instrumentation packages. I am not able to visualize the release process in this scenario where we add separate instrumentation package at the root of this repository.

README.md
```diff
@@ -216,6 +216,49 @@ xray.AWS(dynamo.Client)
...
+
+	// Build the request with its input parameters
+	_, err = svc.ListTables(ctx, &dynamodb.ListTablesInput{
+		Limit: aws.Int32(5),
+	})
+	if err != nil {
+		log.Fatalf("failed to list tables, %v", err)
+	}
+
+	root.Close(nil)
```
callado4, line 258: Is this `root.Close(nil)` needed again since there is already a `defer root.Close(nil)` above?

bhautikpip: It's not. good callout.

README.md
bhautikpip: Done.

xray/aws_v2.go
bhautikpip: Done.

xray/aws_v2.go
anuraaga: I was hoping having just a separate package, without a separate go.mod, would be enough for go to not include unneeded code in a downstream binary. Do you know if this is the case or not?

anuraaga: Actually thinking more don't see a problem with the current PR adding a go.mod. What do you mean about the release process? Isn't it just cutting a tag in this repo?

We can't really move around the existing instrumentation without a major version bump so we'll hold off on that, but we may as well try to split better for new instrumentation.

bhautikpip: Basically, currently we are tagging the entire repository as released version. If we separate instrumentation package then It means some modules in the instrumentation packages relies on core `xray` package. So, I am not able to visualize how we are going to tag these packages with out reworking the SDK. Ideally what we should do is to have separate module for core SDK and separate module for instrumentation and we should follow the release cycle of core SDK and separate instrumentation packages rather than tagging an entire repository.  If we take open telemetry contrib repo as an example here's how it released packages looks like: https://github.com/open-telemetry/opentelemetry-go-contrib/releases . I am not sure about this - https://github.com/aws/aws-xray-sdk-go/pull/309#discussion_r653293298 AFAIK In order to reduce the dependencies it has to be the module but I will dig more and update the thread.

wangzlei: Any concern if we just create separate go.mod for aws sdk v2 instrumentation, not change legacy instrumentations?

bhautikpip: @wangzlei As I have mentioned it would require moving things around within SDK. Right now instrumentation packages and core SDK packages resides in `xray` dir. We probably have to separate out those components. @anuraaga and I had a discussion regarding the same so we end up deciding to have a separate package for aws v2 instrumentation and not a module right now to gain some advantage of dead code elimination. Also, this would require a major version release since this is a breaking change.

anuraaga (approved):

bhautikpip:
Update: I am currently validating the dependency size after adding aws sdk instrumentation dependencies to see if the number is not huge once I can confirm I will merge this PR.

bhautikpip:
@wangzlei tested below code (which has xray and awsv2 imports) to check for binary size and the size was 21.2 MB. I think this instrumentation does not add significant amount of overhead in size so should be good to merge. 

```
package main

import (
	"context"
	"log"

	"github.com/aws/aws-sdk-go-v2/aws"
	"github.com/aws/aws-sdk-go-v2/config"
	"github.com/aws/aws-sdk-go-v2/service/dynamodb"
	"github.com/aws/aws-xray-sdk-go/instrumentation/awsv2"
)

func main() {
	ctx, root := xray.BeginSegment(context.TODO(), "AWSSDKV2_Dynamodb")
	defer root.Close(nil)
	cfg, err := config.LoadDefaultConfig(ctx, config.WithRegion("us-west-2"))
	if err != nil {
		log.Fatalf("unable to load SDK config, %v", err)
	}
	// Instrumenting AWS SDK v2
	awsv2.AWSV2Instrumentor(&cfg.APIOptions)
	// Using the Config value, create the DynamoDB client
	svc := dynamodb.NewFromConfig(cfg)
	// Build the request with its input parameters
	_, err = svc.ListTables(ctx, &dynamodb.ListTablesInput{
		Limit: aws.Int32(5),
	})
	if err != nil {
		log.Fatalf("failed to list tables, %v", err)
	}
}
```

PR #311, opened 2021-06-16
fix: memory leak fasthttp handler

nicolascb:
We can't use `fasthttp.RequestCtx` in `NewSegmentFromHeader`, this generated an infinite wait in [xray/segment.go#L135](https://github.com/aws/aws-xray-sdk-go/blob/7446e854f1b2b9e0f2846cd33045abf4896dc0c8/xray/segment.go#L135), because `fasthttp.RequestCtx.Done()` is not called after finishing the request.

The solution was to create a new context to pass to `NewSegmentFromHeader`.

### DIFF
--- a/xray/fasthttp.go
+++ b/xray/fasthttp.go
@@ -1,6 +1,7 @@
 package xray
 
 import (
+	"context"
 	"fmt"
 	"net/http"
 	"net/url"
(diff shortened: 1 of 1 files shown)


nicolascb:
@bhautikpip We depend on this change to put X-Ray into production, can you help me? :D

bhautikpip:
Sure. taking a look!

bhautikpip (approved):
Looks good :)

Issue #51, opened 2018-04-10
goroutine leaks from beginSegment


elan-sg:
https://github.com/aws/aws-xray-sdk-go/blob/8d5ff7cd4d96b6e380cf7bcaf576ed1848b10b30/xray/segment.go#L70-L75

This goroutine will leak if a context without cancel is used to beginSegment, I think Segment.Close should clean this up.

haotianw465:
Hi, 

Sorry for the late response. Thank you for reporting the issue. We are aware of this and will have a fix soon. Please stay tuned.

Cyberax:
Guys, it's been a year so far. The fix is simple, just add a select{} statement that monitors a channel that is closed in segment.Close().

yogiraj07:
Hello @Cyberax ,
Apologies for delay in solving this issue. I am happy to work with you and prioritize on solving it.  Can you post a PR for the above commit with unit tests. 

Best,
Yogi

LeoAdamek:
Any update on this? It's quite a big issue as, even when I explicitly complete the context passed to the segment, there are still goroutines leaked. In an API the rate these leaks occur can be up to the every request.

In my case I've written a minimal middleware for Gin to handle tracing:

~~~~go
g.Use(func(c *gin.Context) {
		ctx := context.WithValue(c, ctxURL, c.Request.URL.String())

		serviceName := "solos-api"

		if envName := viper.GetString("xray.name"); envName != "" {
			serviceName += " " + envName
		}

		parts := strings.Split(c.HandlerName(), ".")

		serviceName += " " + parts[len(parts)-1]

		ctx, s := xray.BeginSegment(ctx, serviceName)

		remote := c.Request.Header.Get("X-Forwarded-For")

		if remote == "" {
			remote = c.Request.RemoteAddr
		}

		s.HTTP = &xray.HTTPData{}
		s.HTTP.Request = &xray.RequestData{
			Method:    c.Request.Method,
			ClientIP:  remote,
			URL:       c.Request.URL.Path,
			UserAgent: c.Request.UserAgent(),
		}

		// Add the trace Id as a header so the request can be referenced later
		// for troubleshooting.
		c.Writer.Header().Set("X-Trace-Id", s.TraceID)

		c.Set("ctx", ctx)
		c.Next()
		status := c.Writer.Status()
		s.HTTP.Response = &xray.ResponseData{
			Status: status,
		}

		s.Error = false
		s.Fault = false

		if status >= 400 {
[...]

PR #327, opened 2021-08-31
xray: Don't leak goroutines on segment start

narqo:
*Issue #, if available:*
This is a follow-up for https://github.com/aws/aws-xray-sdk-go/issues/51 and https://github.com/aws/aws-xray-sdk-go/pull/156

*Description of changes:*
https://github.com/aws/aws-xray-sdk-go/pull/156 has fixed a leak caused by the use of `context.Background()` with `xray.BeginSegmentWithSampling`. This won't work if a service's background job, uses a context, whose cancellation is bound to service's life-time. E.g. consider the following code:

```go
func main() {
	ctx, stop := signal.NotifyContext(context.Background(), os.Interrupt)
	defer stop()

	ticker := time.NewTicker(time.Minute)
	defer ticket.Stop()
	for {
		select {
		case <-ticker.C:
			_, seg := xray.BeginSegment(ctx, "")  // 👈 spawns a goroutine that never stops
			// do something
			seg.Close(nil)
		case <-ctx.Done():
			break
		}
	}
}
```

This PR aims at fixing the issue by making sure the goroutine, that `xray.BeginSegment` spawns will be stopped after the segment gets closed.

### DIFF
--- a/xray/segment.go
+++ b/xray/segment.go
@@ -130,17 +130,19 @@ func BeginSegmentWithSampling(ctx context.Context, name string, r *http.Request,
 		}
 	}
 
-	if ctx.Done() != nil {
-		go func() {
-			<-ctx.Done()
-			seg.handleContextDone()
-		}()
-	}
-
 	// check whether segment is dummy or not based on sampling decision
 	if !seg.ParentSegment.Sampled {
 		seg.Dummy = true
 	}
+
+	// create a new context to close it on segment close;
+	// this makes sure the closing of the segment won't affect the client's code, that uses the returned context
+	ctx1, cancelCtx := context.WithCancel(ctx)
+	seg.cancelCtx = cancelCtx
+	go func() {
+		<-ctx1.Done()
+		seg.handleContextDone()
+	}()
 
 	// generates segment and trace id based on sampling decision and AWS_XRAY_NOOP_ID env variable
 	idGeneration(seg)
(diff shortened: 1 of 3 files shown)


xray/segment_model.go
```diff
@@ -23,9 +24,10 @@ type Segment struct {
 	parent           *Segment
 	openSegments     int
 	totalSubSegments uint32
-	Sampled          bool           `json:"-"`
-	RequestWasTraced bool           `json:"-"` // Used by xray.RequestWasTraced
-	ContextDone      bool           `json:"-"`
+	Sampled          bool `json:"-"`
```
bhautikpip, line 27: Can we reformat this so that `cancelCtx` would be the only new line?

xray/segment.go
```diff
@@ -130,18 +130,20 @@ func BeginSegmentWithSampling(ctx context.Context, name string, r *http.Request,
...
-	}
-
 	// check whether segment is dummy or not based on sampling decision
 	if !seg.ParentSegment.Sampled {
 		seg.Dummy = true
 	}
 
+	// create a new context to close it on segment close;
+	// this makes sure the closing of the segment won't affect the client's code, that uses the returned context
+	ctx1, cancelCtx := context.WithCancel(ctx)
```
bhautikpip, line 140: Can we write a small unit test to confirm this behavior?

narqo:
🆙 

@bhautikpip I've push the updated version, PHAL

narqo:
@bhautikpip is there anything blocking from merging this PR?

wangzlei (approved):

PR #342, opened 2021-12-17
Add caution to not use the BeginFacadeSegment API

srprash:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/xray/segment.go
+++ b/xray/segment.go
@@ -54,7 +54,7 @@ func noOpSegmentID() string {
 }
 
 // BeginFacadeSegment creates a Segment for a given name and context.
-// Note: This is an internal API only to be used in Lambda context within the SDK. Consider using BeginSegment instead.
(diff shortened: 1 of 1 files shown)


xray/segment.go
```diff
@@ -54,6 +54,7 @@ func noOpSegmentID() string {
 }
 
 // BeginFacadeSegment creates a Segment for a given name and context.
+// Caution: This is an internal API only to be used in Lambda context within the SDK. Consider using BeginSegment instead.
```
bhautikpip, line 57: Nit: Maybe can use `Note` instead of `Caution` ?

bhautikpip (approved):

xray/segment.go
NathanielRN: +1, having `NOTE` specifically usually gets picked up by text editors, i.e. :

image

NathanielRN (approved):

PR #344, opened 2022-01-18
Remove dep pkg manager and excess CI steps

### LINKED ISSUE #266: Remove support for dep
go modules have been stable for a long time and now widely used in the ecosystem. Our tooling will become much simpler if we go ahead and remove support for dep, for example #265 becomes trivial to solve by compressing our build down to a single command.

NathanielRN:
## Description

Thanks for bringing this issue up @anuraaga ! I looked through old PRs like #90 and read your description to come up with this PR to remove `dep`.

~One concern I have was that you said we could~

~> for example #265 becomes trivial to solve by compressing our build down to a single command.~

~so I looked at how we ended up fixing #265 in #267 and noticed [these lines](https://github.com/aws/aws-xray-sdk-go/pull/267/files#diff-faff1af3d8ff408964a57b2e475f69a6b7c7b71c9978cccc8f471798caac2c88R34-R35):~

```go
go version
echo "GOPATH=${{ github.workspace }}" >> $GITHUB_ENV
echo "${{ github.workspace }}/bin" >> $GITHUB_PATH
```

~I'm not too familiar with `go`, but does removing `dep` mean we don't need these lines anymore?~

~Actually I just [tested this on my own branch on my fork](https://github.com/NathanielRN/aws-xray-sdk-go/runs/4860035736?check_suite_focus=true) and it works fine so I will remove these lines. Please let me know if I'm wrong!~

~Also when you say~

> by compressing our build down to a single command.

~Did I do that correctly in this PR because we only have `go help download` now?~

~If it would be more convenient to link me to `dep` documentation where I can verify this is not needed please feel free to do that!~

Did a final test with all your comments and the test still passes [on my fork](https://github.com/NathanielRN/aws-xray-sdk-go/actions/runs/1719812674). This change should drop the size of the CI a lot!

Fixes #266 


.github/workflows/test.yml
```diff
@@ -30,15 +30,13 @@ jobs:
         with:
           go-version: ${{ matrix.go }}
         id: go
-      - name: Set up GOPATH
+      - name: Set up INSTALL_DIRECTORY
         run: |
-          go version
-          echo "GOPATH=${{ github.workspace }}" >> $GITHUB_ENV
-          echo "${{ github.workspace }}/bin" >> $GITHUB_PATH
+
```
NathanielRN, line 35: I left this space because my VSCode IDE only does correct syntax highlighting for the comment if I have this space here... let me know if this is undesirable!

.github/workflows/test.yml
```diff
@@ -30,15 +30,13 @@ jobs:
...
           mkdir -p "bin"
           mkdir -p "pkg"
           mkdir -p "src"
```
anuraaga, line 38: Maybe for another PR but it looks like these should be handled by the makefile (well I can't imagine test-with-race to need any folders lke this set up)

.github/workflows/test.yml
```diff
@@ -48,12 +46,7 @@ jobs:
 
       - name: Install Dependencies
         run: |
-          if go help mod > /dev/null 2>&1 ; then
-            go mod download
-          else
-            curl https://raw.githubusercontent.com/golang/dep/master/install.sh | sh
-            dep ensure
-          fi
+          go mod download
```
anuraaga, line 49: `go mod download` tends to download much more than needed, especially if not actively `go mod tidy`ing the mod files. It's fine to remove this and let Go automatically download what's needed

.github/workflows/test.yml
```diff
@@ -48,12 +46,7 @@ jobs:
...
         env:
           GO111MODULE: "on"
```
anuraaga, line 51: Don't need this env setting for 1.16+ which is what this build targets

.github/workflows/test.yml
NathanielRN: Thanks for the call out! You're absolutely right, I removed these steps and ran them on my fork. I found out that [the tests pass just fine without them](https://github.com/NathanielRN/aws-xray-sdk-go/actions/runs/1719812674).

I thought about making another PR but since this is just deleting lines I thought it would be fine to put them in the same PR. I'll just update the PR title!

.github/workflows/test.yml
NathanielRN: Great sounds good. Removed!

.github/workflows/test.yml
NathanielRN: Good call out, thanks! I've removed it.

anuraaga (approved):

Issue #352, opened 2022-03-30
Consider using `errors.As` instead of type assertions in `segment.Close` to support wrapped errors


Okonos:
`Segment.Close` enriches the segment with details specific to the error passed, but since it uses type assertions, the error must be of an exact type:

https://github.com/aws/aws-xray-sdk-go/blob/053264401e5a9608d10a4c44a613cacd607cfe94/strategy/exception/default_exception_formatting_strategy.go#L118-L123
https://github.com/aws/aws-xray-sdk-go/blob/053264401e5a9608d10a4c44a613cacd607cfe94/strategy/exception/default_exception_formatting_strategy.go#L136-L138
https://github.com/aws/aws-xray-sdk-go/blob/053264401e5a9608d10a4c44a613cacd607cfe94/strategy/exception/default_exception_formatting_strategy.go#L143-L145

Using [`errors.As`](https://pkg.go.dev/errors#As) there instead of type assertions would make it work with wrapped errors as well.

PR #353, opened 2022-03-31
Replace error type assertions with `errors.As` in DefaultFormattingStrategy

Okonos:
*Issue #352*

*Description of changes:*
See the linked issue.

### DIFF
--- a/strategy/exception/default_exception_formatting_strategy.go
+++ b/strategy/exception/default_exception_formatting_strategy.go
@@ -11,6 +11,7 @@ package exception
 import (
 	"bytes"
 	"crypto/rand"
+	goerrors "errors"
 	"fmt"
 	"runtime"
 	"strings"
@@ -115,7 +116,8 @@ func (dEFS *DefaultFormattingStrategy) Panicf(formatString string, args ...inter
 // ExceptionFromError takes an error and returns value of Exception
 func (dEFS *DefaultFormattingStrategy) ExceptionFromError(err error) Exception {
 	var isRemote bool
-	if reqErr, ok := err.(awserr.RequestFailure); ok {
+	var reqErr awserr.RequestFailure
+	if goerrors.As(err, &reqErr) {
 		// A service error occurs
 		if reqErr.RequestID() != "" {
 			isRemote = true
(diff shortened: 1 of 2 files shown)


strategy/exception/default_exception_formatting_strategy.go
```diff
@@ -133,22 +135,25 @@ func (dEFS *DefaultFormattingStrategy) ExceptionFromError(err error) Exception {
 		Remote:  isRemote,
 	}
 
-	if err, ok := err.(*XRayError); ok {
-		e.Type = err.Type
+	xRayErr := &XRayError{}
+	if goerrors.As(err, &xRayErr) {
```
bhautikpip, line 139: not sure what is the reasoning behind this, but can we do it like below?

```
if goerrors.As(err, &XRayError{}) {
		e.Type = xRayErr.Type
}
```
or 
```
var xRayErr XRayError
if goerrors.As(err, &xRayErr}) {
		e.Type = xRayErr.Type
}
```

strategy/exception/default_exception_formatting_strategy.go
```diff
@@ -115,7 +116,8 @@ func (dEFS *DefaultFormattingStrategy) Panicf(formatString string, args ...inter
 // ExceptionFromError takes an error and returns value of Exception
 func (dEFS *DefaultFormattingStrategy) ExceptionFromError(err error) Exception {
 	var isRemote bool
-	if reqErr, ok := err.(awserr.RequestFailure); ok {
+	var reqErr awserr.RequestFailure
```
bhautikpip, line 119: Also, I think in general it would be a good idea to add some test cases around this to confirm the behavior.

Okonos: Sure, I've added some basic cases

strategy/exception/default_exception_formatting_strategy.go
Okonos: The `err` was previously casted to a pointer type (`*XRayError`), so to preserve that behavior an address of a pointer is passed down to `errors.As`.
The first suggestion seems to be incorrect (`xRayErr` variable undefined), the second leaves us with a value, not a pointer.

bhautikpip (approved):
LGTM :)

PR #362, opened 2022-05-10
fix: Memory leakage case of dummy segment

sawadashota:
*Description of changes:*

#327 causes memory leakage when dummy segment. So this PR fix it.

### DIFF
--- a/xray/segment.go
+++ b/xray/segment.go
@@ -357,10 +357,11 @@ func (seg *Segment) Close(err error) {
 		seg.addError(err)
 	}
 
+	cancelSegCtx := seg.cancelCtx
 	defer func() {
 		// makes sure the goroutine, that waits for possible top-level context cancellation gets closed on segment close
-		if seg.cancelCtx != nil {
-			seg.cancelCtx()
+		if cancelSegCtx != nil {
+			cancelSegCtx()
 		}
 	}()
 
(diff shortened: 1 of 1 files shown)


nikolaianohyn (approved):

mathn00b (approved):

prigornitskiy (approved):

xray/segment.go
```diff
@@ -357,21 +357,22 @@ func (seg *Segment) Close(err error) {
 		seg.addError(err)
 	}
 
+	cancelSegCtx := seg.cancelCtx
```
nikolaianohyn, line 360: another way: don't create routine for Sampled segments

```
	// create a new context to close it on segment close;
	// this makes sure the closing of the segment won't affect the client's code, that uses the returned context
	if !seg.Dummy {
		ctx1, cancelCtx := context.WithCancel(ctx)
		seg.cancelCtx = cancelCtx
		go func() {
			<-ctx1.Done()
			seg.handleContextDone()
		}()
	}
```

nikolaianohyn: For me it's better because you don't send sampled segments and don't create extra go routines

NathanielRN (commented):
Thanks for this change! I had some questions but this is looking good!

xray/segment.go
nikolaianohyn: Hi @NathanielRN [here PR](https://github.com/aws/aws-xray-sdk-go/pull/365)

sawadashota: Thanks @nikolaianohyn ! I think your PR is better so I will close this PR.

Issue #364, opened 2022-05-18
Memory leak with BeginSubsegment


thinhlvv:
I'm getting memory leak issue when executing BeginSubSegment.
Here is the profiling graph we got when the issue was happening.

AWS X-Ray SDK for Go: v1.7.0
Go version: go1.16.2 darwin/amd64

![Screenshot 2022-05-18 at 3 44 31 PM](https://user-images.githubusercontent.com/14247456/168987222-d3055339-d77f-405e-b20d-04c3b3cf8c77.png)
![Screenshot 2022-05-18 at 3 47 08 PM](https://user-images.githubusercontent.com/14247456/168987230-41205137-291c-4d0e-a1ea-e4681803fe21.png)

nikolaianohyn:
For sampled segments Close() function doesn't cancel context (!)
`// check whether segment is dummy or not based on sampling decision
	if !seg.ParentSegment.Sampled {
		seg.Dummy = true
	}

	// create a new context to close it on segment close;
	// this makes sure the closing of the segment won't affect the client's code, that uses the returned context
	ctx1, cancelCtx := context.WithCancel(ctx)
	seg.cancelCtx = cancelCtx
	go func() {
		<-ctx1.Done()
		seg.handleContextDone()
	}()`
	
Fragment from close function:
`// If segment is dummy we return
	if seg.Dummy {
		seg.Unlock()
		return
	}

	cancelSegCtx := seg.cancelCtx

	seg.Unlock()
	seg.send()

	// makes sure the goroutine, that waits for possible top-level context cancellation gets closed on segment close
	if cancelSegCtx != nil {
		cancelSegCtx()
	}`	
	
	So if segment "Dummy" we have leak routine

prigornitskiy:
fix: https://github.com/aws/aws-xray-sdk-go/pull/362

stijndehaes:
We are seeing memory leaks in production for one of our applications. Could these dummy segments be the culprit? I see most memory usage is in Xray for the moment, still need to investigate it further. If so could we perhaps have a release with the PR with the fix?

willarmiros:
Hi @stijndehaes, thanks for following up. We will be cutting a release as soon as able to get this fix out.

stijndehaes:
Hi @willarmiros, thank you very much! I yesterday just upgraded to the commit with the fix on main and saw immediate results, memory pressure was way down, so I can confirm it helps!

byrnedo:
[...]

PR #365, opened 2022-05-18
fix: routine leak for dummy segments

nikolaianohyn:
Issue [364](https://github.com/aws/aws-xray-sdk-go/issues/364)

 If segment is dummy don't create go routine for context cancelation handling

### DIFF
--- a/xray/segment.go
+++ b/xray/segment.go
@@ -136,14 +136,17 @@ func BeginSegmentWithSampling(ctx context.Context, name string, r *http.Request,
 		seg.Dummy = true
 	}
 
-	// create a new context to close it on segment close;
-	// this makes sure the closing of the segment won't affect the client's code, that uses the returned context
-	ctx1, cancelCtx := context.WithCancel(ctx)
-	seg.cancelCtx = cancelCtx
-	go func() {
-		<-ctx1.Done()
-		seg.handleContextDone()
-	}()
+	// don't allocate resources for dummy segments
+	if !seg.Dummy {
+		// create a new context to close it on segment close;
+		// this makes sure the closing of the segment won't affect the client's code, that uses the returned context
+		ctx1, cancelCtx := context.WithCancel(ctx)
+		seg.cancelCtx = cancelCtx
+		go func() {
+			<-ctx1.Done()
+			seg.handleContextDone()
+		}()
+	}
 
 	// generates segment and trace id based on sampling decision and AWS_XRAY_NOOP_ID env variable
 	idGeneration(seg)


prigornitskiy (approved):

NathanielRN:
Thanks for this PR @nikolaianohyn! I agree that this should fix it 🙂

However, can we still incorporate @sawadashota's solution here? I think it's good to be safe even if we don't add the `cancelCtx` to the segment anymore.

i.e. changing this code:

https://github.com/aws/aws-xray-sdk-go/blob/b4b13727e8c47e1c3227958df028be4cd823d3d7/xray/segment.go#L360-L374

To look like this:

```go
cancelSegCtx := seg.cancelCtx

seg.Unlock()

// If a goroutine was started to close the extra context (made for
// cancelling the segment), stop the goroutine before we close the segment
// and lose access to it.
if cancelSegCtx != nil {
	cancelSegCtx()
}

// If segment is dummy we return
if seg.Dummy {
	return
}

seg.send()
```

I don't see any precedence on testing Dummy segments so I won't ask for it in this PR, but if you were interested in trying to make one I'd be happy to review it 😄

xray/segment.go
```diff
@@ -136,14 +136,17 @@ func BeginSegmentWithSampling(ctx context.Context, name string, r *http.Request,
...
-	ctx1, cancelCtx := context.WithCancel(ctx)
-	seg.cancelCtx = cancelCtx
-	go func() {
-		<-ctx1.Done()
-		seg.handleContextDone()
-	}()
+	// don't allocate resources for dummy segments
+	if !seg.Dummy {
+		// create a new context to close it on segment close;
+		// this makes sure the closing of the segment won't affect the client's code, that uses the returned context
```
NathanielRN, line 142: I think we can improve on the original wording, Does this makes sense to you? Let me know!

```suggestion
	// Dummy segments don't get sent and don't need a goroutine to cancel them.
	if !seg.Dummy {
		// Create a new context for to cancel segment.
		// Start new goroutine to listen for segment close/cancel events.
		// Cancel simply calls `segment.Close()`.
		//
		// This way, even if the client consumes the `ctx.Done()` event, the
		// X-Ray SDK has a way of canceling (closing) the segment in the
		// `segment.Close()` method using this new cancellation context.
```

NathanielRN:
The unit tests are failing but I don't believe the changes from this PR are related. We've seen tests fail on re-run [because of global state before](https://github.com/aws/aws-xray-sdk-go/pull/327#issuecomment-941282708) so I'll merge this PR despite the failing tests because the have passed before with [the original change](https://github.com/aws/aws-xray-sdk-go/actions/runs/2347532173).

PR #375, opened 2022-08-11
Local test daemon for grpc server test cases

srprash:
*Issue:*
For the last few runs of the unit test CI, the [`TestUnaryServerInterceptor`](https://github.com/aws/aws-xray-sdk-go/blob/0f2ea34563faaf968d497506bd21f71515af42af/xray/grpc_test.go#L256) has been failing due to inconsistent segment across test cases.

*Description of changes:*
The cause for above flaky test cases failure is the shared channel in the `TestDaemon` for the test cases. I tested through a separate experiment that a `channel` in Go is not guaranteed to return the same value in a goroutine which was put there by the same goroutine. 
In [the test cases](https://github.com/aws/aws-xray-sdk-go/blob/0f2ea34563faaf968d497506bd21f71515af42af/xray/grpc_test.go#L270), there are 4 goroutines created (when the test runs with `-race` flag) which create individual segments but retrieve the segment through the common test daemon channel. This leads to a segment of one test case leaking into another.

To avoid segment mismatch across test cases, we can have a local instance of test daemon for each test case. 

### DIFF
--- a/xray/grpc_test.go
+++ b/xray/grpc_test.go
@@ -254,19 +254,6 @@ func TestGrpcUnaryClientInterceptor(t *testing.T) {
 }
 
 func TestUnaryServerInterceptor(t *testing.T) {
-	ctx, td := NewTestDaemon()
-	defer td.Close()
-
-	lis := newGrpcServer(
-		t,
-		grpc.UnaryInterceptor(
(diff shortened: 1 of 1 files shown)


wangzlei (approved):
LGTM!

Issue #334, opened 2021-10-29
Feature request: Don't panic when a subsegment can't be found


a-h:
When I run this code on my local machine, I don't expect a panic.

```go
package main

import (
	"fmt"
	"io"
	"os"

	"github.com/aws/aws-xray-sdk-go/xray"
)

func main() {
	c := xray.Client(nil)
	resp, err := c.Get("https://google.com")
	if err != nil {
		fmt.Printf("failed to access Google: %v\n", err)
		os.Exit(1)
	}
	io.Copy(os.Stdout, resp.Body)
}
```

But it panics!

```
panic: failed to begin subsegment named 'google.com': segment cannot be found.
```

I was hunting around in the docs to find out how to suppress this behaviour. I'd like my code to use X-Ray, if it's available, but if it's not, I don't care.

I found that I can set an environment variable, but that's a pain, because I'd even have to set that environment variable to run some unit tests or integration tests locally.

```
AWS_XRAY_CONTEXT_MISSING=IGNORE_ERROR
```

As per the Go Proverbs, don't panic please. 😁 

Suggestions:

* Change the default behaviour so that it doesn't panic because this is unexpected. If you really *must* panic, explain in the panic message how to fix or disable it.
* Make it more obvious that this would happen, and provide a way to turn off this behaviour without an environment variable (or document it if there is).

bhautikpip:
Hey @a-h,

Looks like panic is expected in your code because you're trying to create a subsegment without creating a segment first. Ideally either you should create manual segment using `xray.BeginSegment` API or maybe wrap your HTTP client around HTTP handler instrumentation (examples can be found in README) then X-Ray Go SDK will automatically creates a segment and then Go SDK will create HTTP subsegment. Yeah you're right you can avoid panic using `IGNORE_ERROR` option which would just ignore the panic and continue execution of your code.

a-h:
Thanks @bhautikpip, I understand why it happens, that's not the point of the issue.

I'm saying that it's a footgun that the default behaviour of the library is to panic over nothing.

The current default behaviour is not expected, see https://go.dev/blog/defer-panic-and-recover and https://go.dev/blog/errors-are-values

> The convention in the Go libraries is that even when a package uses panic internally, its external API still presents explicit error return values.

i.e. the API should return an error value (or do nothing) rather than panic. It's total overkill to crash the program just because an XRay subsegment can't be created.

bhautikpip:
[...]

PR #382, opened 2022-10-21
Setting Default ErrorStrategy to Log

atshaw43:
*Issue #, if available:*

*Description of changes:*
Setting Default ErrorStrategy to Log

### DIFF
--- a/xray/segment.go
+++ b/xray/segment.go
@@ -352,6 +352,11 @@ func (seg *Segment) Close(err error) {
 		return
 	}
 
+	if seg == nil {
+		logger.Debugf("No input subsegment to end. No-op")
+		return
+	}
+
 	seg.Lock()
 	if seg.parent != nil {
 		logger.Debugf("Closing subsegment named %s", seg.Name)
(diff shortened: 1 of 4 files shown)


srprash:
@atshaw43 Will this change address this request? https://github.com/aws/aws-xray-sdk-go/issues/334

xray/segment.go
```diff
@@ -352,6 +352,11 @@ func (seg *Segment) Close(err error) {
 		return
 	}
 
+	if seg == nil {
+		logger.Debugf("No input subsegment to end. No-op")
```
srprash, line 356: Is there a way we can delegate this situation to be handled by the ContextMissingStrategy?
This way it can be just a log or a runtime exception if a user choses to.

atshaw43: I talked with Will in regards to its implementation in Java.

"I think that’s the expected behavior for this method. If a customer explicitly passes in the subsegment to be closed, then we shouldn’t throw context missing errors because we’re not interacting with the context, we just use the subsegment provided by the customer as noted in [javadocs](https://github.com/aws/aws-xray-sdk-java/blob/fd93c8a938e635137dedc70edf9b709f039ba0c1/aws-xray-recorder-sdk-core/src/main/java/com/amazonaws/xray/AWSXRayRecorder.java#L567-L568)."

srprash (approved):
Thanks!

PR #384, opened 2022-11-11
Fixing sample app

atshaw43:
*Issue #, if available:*

*Description of changes:*
The sample app used in integration tests needs a dependency on the aws lambda library to run.


sample-apps/http-server/go.sum
```diff
@@ -4,6 +4,7 @@ github.com/DATA-DOG/go-sqlmock v1.4.1 h1:ThlnYciV1iM/V0OSF/dtkqWb6xo5qITT1TJBG1M
 github.com/DATA-DOG/go-sqlmock v1.4.1/go.mod h1:f/Ixk793poVmq4qj/V1dPUg2JEAKC73Q5eFN3EC/SaM=
 github.com/andybalholm/brotli v1.0.4 h1:V7DdXeJtZscaqfNuAdSRuRFzuiKlHSC/Zh3zl9qY3JY=
 github.com/andybalholm/brotli v1.0.4/go.mod h1:fO7iG3H7G2nSZ7m0zPUDn85XEX2GTukHGRSepvi9Eig=
+github.com/aws/aws-lambda-go v1.34.1 h1:M3a/uFYBjii+tDcOJ0wL/WyFi2550FHoECdPf27zvOs=
```
willarmiros, line 7: Should we be updating the `go.mod` file instead?

atshaw43: If I edit it in the mod file, then it rips its out during "go mod tidy". I think it is because project itself does not have a direct dependency. It depends on a project that depends on a it.

atshaw43: When I ran "go mod download github.com/aws/aws-lambda-go" on the machine, then it worked.
So I ran it locally (after moving a copy of the SDK into the same folder) and this is the change.

willarmiros: I see, it's a transitive dependency of the SDK. Makes sense thanks for looking into this!

willarmiros (approved):
LGTM

PR #387, opened 2022-11-12
Fixing Sample App 

atshaw43:
*Issue #, if available:*

*Description of changes:*
Downloads the AWS lambda module in the sample app.


sample-apps/http-server/Dockerfile
```diff
@@ -5,6 +5,7 @@ COPY . .
 ENV GOPROXY direct
 
 RUN go mod download
+RUN go mod download github.com/aws/aws-lambda-go
```
willarmiros, line 8: Hmm this seems strange, why don't we have to do this for any of our other dependencies?

More concerning, would this change also break our customer's apps if they don't already have this dependency?

atshaw43: Yea, I am not sure why this dependency is different. I'll take another look.

PR #389, opened 2022-11-17
Update README.md

atshaw43:
*Issue #, if available:*

*Description of changes:*
Adding oversampling mitigation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -329,7 +329,7 @@ func HandleRequest(ctx context.Context, name string) (string, error) {
 ```
 
 **Oversampling Mitigation**
-Oversampling mitigation allows you to ignore a parent segment/subsegment's sampled flag and instead set it to false.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -328,6 +328,50 @@ func HandleRequest(ctx context.Context, name string) (string, error) {
 }
 ```
 
+**Oversampling Mitigation**
+Oversampling mitigation allows you to ignore a parent segment/subsegment's sampled flag and instead set it to false.
```
carolabadeer, line 332: Can this explanation of oversampling mitigation be reworded to be more clear? I think it is important to clarify what _it_ refers to in "instead set it to false" and also explain how the subsegments now inherit the sampling decision of their direct parent - segment or subsegment -  as opposed to always referring to the parent segment.

carolabadeer (approved):

PR #396, opened 2023-02-22
Change workflow to use go v1.9 rather than v1

jj22ee:
*Issue #, if available:*
Workflow unit tests are failing in environments (windows-latest, 1) and (macos-latest, 1).

*Description of changes:*
Since the AWS X-Ray SDK for Go is compatible with Go v1.9 and above, remove the unit tests for Go v1 and use Go v1.9

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -21,7 +21,7 @@ jobs:
         go:
           - '1.16'
           - '1.17'
-          - '1'
+          - '1.9'
     steps:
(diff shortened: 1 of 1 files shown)


willarmiros:
I think the Readme is outdated, Go v1.9 has been EOL for years. We just haven't updated the Go version in the readme as we've updated the supported versions.

"Go v1" simply tests the latest minor version of Go, we should keep that as part of the CI to ensure we're running against the latest version and fix any breaking changes introduced.

PR #449, opened 2023-11-14
Changing sample-app go.mod redirection

atshaw43:
Issue #, if available:
The sample-app does not work out of the box because the go.mod redirection points to the wrong folder.

Description of changes:
Change the redirection to 2 folders up instead. This causes an issue with EB running with Go. So we change it back for the GitHub workflows.

Other option tried:
I tried changing the folder structure of deploy.zip to get this to work without changing the text of go.mod, but this does not seem possible given the limitations of EB. This would have been the preferred option to me because I find editing code/config to to test to be higher maintenance. However, application.go must be at the working directory level. This is hard coded as a part of EB. And it is awkward to try to place the aws-xray-sdk-go folder two levels up.
https://docs.aws.amazon.com/elasticbeanstalk/latest/dg/go-environment.html


wangzlei (approved):
LGTM!

PR #419, opened 2023-09-27
suppress Panic in Emitter

wangzlei:
*Issue #, if available:*

The potential concurrency risk could lead to Default_Emitter.Emit() Panic during JSON marshaling. We would like to use recover() to prevent X-ray SDK crashes in the user's application before addressing the root cause.

* fatal error 1:

```
panic: runtime error: index out of range [13] with length 13 [recovered]
    panic: runtime error: index out of range [13] with length 13

goroutine 41146499 [running]:
encoding/json.(*encodeState).marshal.func1()
    /usr/local/go/src/encoding/json/encode.go:291 +0x80
panic({0x1162f60?, 0x4001aa78a8?})
    /usr/local/go/src/runtime/panic.go:914 +0x218
encoding/json.mapEncoder.encode({0x40031e77a8?}, 0x40044c8340, {0x102d760?, 0x4001410478?, 0xffffffffffffffff?}, {0xe?, 0x0?})
    /usr/local/go/src/encoding/json/encode.go:745 +0x5b0
encoding/json.structEncoder.encode({{{0x4002bd6000, 0x19, 0x23}, 0x4000521d10, 0x4000521ec0}}, 0x40044c8340, {0x1205d60?, 0x4001410340?, 0x4002bd6000?}, {0x19?, ...})
    /usr/local/go/src/encoding/json/encode.go:706 +0x190
encoding/json.ptrEncoder.encode({0x40031e79c8?}, 0x40044c8340, {0x11f52e0?, 0x4001410340?, 0x40031e79d8?}, {0x60?, 0x72?})
    /usr/local/go/src/encoding/json/encode.go:878 +0x1c4
encoding/json.(*encodeState).reflectValue(0x40031e7a78?, {0x11f52e0?, 0x4001410340?, 0x1258e0?}, {0x40?, 0x65?})
    /usr/local/go/src/encoding/json/encode.go:323 +0x70
encoding/json.(*encodeState).marshal(0x40031e7b28?, {0x11f52e0?, 0x4001410340?}, {0xb4?, 0x8?})
    /usr/local/go/src/encoding/json/encode.go:295 +0xb8
encoding/json.Marshal({0x11f52e0, 0x4001410340})
    /usr/local/go/src/encoding/json/encode.go:162 +0x94
github.com/aws/aws-xray-sdk-go/xray.(*DefaultStreamingStrategy).StreamCompletedSubsegments(0x40031e7d38?, 0x4001410000)
[...]

PR #457, opened 2024-03-12
fix fatal error: concurrent map iteration and map write

wangzlei:
*Issue #, if available:*

If user config StreamingStrategy to be a small value and create many AWS SDK v2 subsegments, 2 goroutines would update the Segment at the same time, causes either `panic: runtime error: index out of range` or `fatal error: concurrent map iteration and map write`.

The panic has been addressed in PR https://github.com/aws/aws-xray-sdk-go/pull/419
The current PR is for addressing the fatal error.

*Description of changes:*


mxiamxia (approved):
LGTM! ty

